PR #1, opened 2016-04-15
CRYPTO-2: Create benchmark for Apache Commons Crypto

winningsix:


.gitignore
```diff
@@ -33,4 +33,5 @@ dist/
 unit-tests.log
 /lib/
 src/main/resources/com/intel/chimera/native/
-src/main/resources/com/intel/chimera/native/*
\ No newline at end of file
+src/main/resources/com/intel/chimera/native/*
```
sundapeng: Change to apache commons

pom.xml
```diff
@@ -331,6 +332,22 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
```
sundapeng: Why we need shade? used for packaging the dependence?

conf/benchmark.properties.template
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+warmupIterations    1000
+iterations          1000
+dataSize            1073741824
+operationSize       8192
+
+bufferSize          20
+cipherClasses       com.intel.chimera.cipher.OpensslCipher,com.intel.chimera.cipher.JceCipher
```
sundapeng: chimera -> crypto

src/test/java/com/intel/chimera/benchmark/ChimeraBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+package com.intel.chimera.benchmark;
+
+import com.intel.chimera.benchmark.option.BenchmarkOption;
+import com.intel.chimera.benchmark.option.StreamOption;
+import com.intel.chimera.cipher.CipherTransformation;
+
+import java.io.*;
+import java.util.Properties;
+
+public class ChimeraBenchmark {
```
sundapeng: ChimeraBenchmark -> CryptoBenchmark

pom.xml
winningsix: We need to add a shade jar to include dependencies used by Common Crypto.

src/test/java/org/apache/commons/crypto/benchmark/CryptoStreamBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+  CryptoInputStream cryptoInputStream;
+  OutputStream cryptoOutputStream;
+  ByteArrayInputStream inputStream;
+  ByteArrayOutputStream outputStream;
+
+  public CryptoStreamBenchmark(BenchmarkOption benchmarkOption,
+      StreamOption streamOption) {
+    this.benchmarkOption = benchmarkOption;
+    this.streamOption = streamOption;
+    outputStream = new ByteArrayOutputStream(benchmarkOption.dataSize);
```
jerrychenhf: It would be better add an little extra space in case that the output size is larger than the input size.  2 * 16 (block size) is enough I think.

src/test/java/org/apache/commons/crypto/benchmark/CryptoStreamBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.Properties;
+import java.util.Random;
+import java.util.TimeZone;
+
+public class CryptoStreamBenchmark {
+  protected String benchMarkName;
+  protected BenchmarkOption benchmarkOption;
+  protected StreamOption streamOption;
+  byte[] inputData;
+  CryptoInputStream cryptoInputStream;
```
jerrychenhf: I suggest not to use the member for the following variables. Localize it will help simplicity of the program.
CryptoInputStream cryptoInputStream; 
OutputStream cryptoOutputStream;  
ByteArrayInputStream inputStream;  
ByteArrayOutputStream outputStream;

src/test/java/org/apache/commons/crypto/benchmark/CryptoStreamBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+  public void testDecryption(int iterations) {
+    try {
+      InputStream cryptoInputStream = getCryptoInputStream();
+
+      byte[] outputData = new byte[benchmarkOption.dataSize];
+
+      System.out.println("Warming up.");
+      // warm up
+      for (int i = 0; i < benchmarkOption.warmupIterations; i++) {
+        doReadOperation(inputStream, cryptoInputStream, outputData);
```
jerrychenhf: We should use the right data for decryption, because for some modes, decryption the random data will throw exceptions.

sbin/run-benchmark.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "This benchmark will evaluate the performance of Chimera in different transfomations, ciphers and JDK versions"
+
+if [ ! -f "conf/benchmark.properties" ];then
+  echo "Not able to find the benchmark.propety, will use default propety instead"
+  cp conf/benchmark.properties.template conf/benchmark.properties
+fi
+
+CRYPTO_JAR=`find . -name commons-crypto*.jar`
+
+echo "Using JDK in path $1 to evalue the performance"
```
jerrychenhf: Suggest to pass singe JDK as the script parameter. People can run twice with different path for benchmarking for different JDK.

PR #2, opened 2016-04-15
CRYPTO-3: Change default cipher as OpensslCipher

winningsix:
This fixes [Change default cipher as OpensslCipher](https://issues.apache.org/jira/browse/CRYPTO-3).

### DIFF
--- a/src/main/java/org/apache/commons/crypto/cipher/CipherFactory.java
+++ b/src/main/java/org/apache/commons/crypto/cipher/CipherFactory.java
@@ -77,17 +77,15 @@ public class CipherFactory {
    */
   public static Cipher getInstance(CipherTransformation transformation)
       throws GeneralSecurityException {
-    return getInstance(transformation, null);
+    return getInstance(transformation, new Properties());
   }
 
+  // Return OpenSSLCipher if Properties is null or empty by default
   private static List<Class<? extends Cipher>> getCipherClasses(Properties props) {
     List<Class<? extends Cipher>> result = new ArrayList<Class<? extends
         Cipher>>();
     String cipherClassString = Utils.getCipherClassString(props);
-    if (cipherClassString == null) {
-      LOG.debug("No cipher classes configured.");
-      return null;
-    }
+
     for (String c : Utils.splitClassNames(cipherClassString, ",")) {
       try {
         Class<?> cls = ReflectionUtils.getClassByName(c);
--- a/src/main/java/org/apache/commons/crypto/conf/ConfigurationKeys.java
+++ b/src/main/java/org/apache/commons/crypto/conf/ConfigurationKeys.java
@@ -17,7 +17,7 @@
  */
 package org.apache.commons.crypto.conf;
 
-import org.apache.commons.crypto.cipher.JceCipher;
+import org.apache.commons.crypto.cipher.OpensslCipher;
 
 /**
  * The ConfigurationKeys contains Configuration keys and default values.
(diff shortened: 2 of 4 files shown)


src/main/java/com/intel/chimera/cipher/CipherFactory.java
```diff
@@ -65,7 +65,7 @@ public static Cipher getInstance(CipherTransformation transformation,
       }
     }
 
-    return (cipher == null) ? new JceCipher(props, transformation) : cipher;
+    return (cipher == null) ? new OpensslCipher(props, transformation) : cipher;
```
jerrychenhf: I would consider this change is dangerous.
Change default cipher to OpensslCipher makes sense but change to OpensslCipher when no Cipher is working may be dangerous. I would suggest, we still use JceCipher here.

src/main/java/com/intel/chimera/cipher/CipherFactory.java
```diff
@@ -77,7 +77,7 @@ public static Cipher getInstance(CipherTransformation transformation,
    */
   public static Cipher getInstance(CipherTransformation transformation)
       throws GeneralSecurityException {
-    return getInstance(transformation, null);
+    return getInstance(transformation, new Properties());
```
jerrychenhf: Is this change needed?

src/main/java/com/intel/chimera/utils/Utils.java
```diff
@@ -135,8 +135,9 @@ public static int getBufferSize(Properties props) {
    */
   public static String getCipherClassString(Properties props) {
     final String configName = CHIMERA_CRYPTO_CIPHER_CLASSES_KEY;
-    return props.getProperty(configName) != null ? props.getProperty(configName) : System
-        .getProperty(configName, CHIMERA_CRYPTO_CIPHER_CLASSES_DEFAULT);
+    return props.getProperty(configName) != null ? props
```
jerrychenhf: If we do want to handle the case that if am empty class string is specified, we want to default to OpensslCipher,
We can check empty string here and set to CLASSES_DEFAULT for empty value.

winningsix:
Thanks  @jerrychenhf  for the reviews. Patch is updated addressing your comments.

src/main/java/org/apache/commons/crypto/cipher/CipherFactory.java
```diff
@@ -47,6 +50,9 @@
    */
   public static Cipher getInstance(CipherTransformation transformation,
       Properties props) throws GeneralSecurityException {
+    if(props == null){
```
jerrychenhf: I would suggest to put the check of props == null into Utils.getCipherClasses method. 
This can centralize the point of handling null props and avoid the need of creating a new Properties object.

src/main/java/org/apache/commons/crypto/cipher/CipherFactory.java
```diff
@@ -80,14 +86,15 @@ public static Cipher getInstance(CipherTransformation transformation)
...
 
+  // Return OpenSSLCipher if Properties is null or empty by default
   private static List<Class<? extends Cipher>> getCipherClasses(Properties props) {
     List<Class<? extends Cipher>> result = new ArrayList<Class<? extends
         Cipher>>();
     String cipherClassString = Utils.getCipherClassString(props);
-    if (cipherClassString == null) {
-      LOG.debug("No cipher classes configured.");
-      return null;
+    if (cipherClassString.isEmpty()) {
```
jerrychenhf: Suggest to put this to centralized place: Utils.getCipherClassString.

winningsix:
Hi @jerrychenhf, I have updated this PR. Please help me review it. Thank you!

jerrychenhf:
+1

winningsix:
Thanks @jerrychenhf for your review. Committed to master at c0df815

PR #3, opened 2016-04-15
[CRYPTO-6] Sync master with Chimera

sundapeng:
https://issues.apache.org/jira/browse/CRYPTO-6 Sync master with the master of Chimera


sundapeng:
Since Chimera have been move to apache commons, it is better for each commit has a JIRA and review in Apache side.

PR #9, opened 2016-04-25
CRYPTO-25: Add download_crypto.cgi to exclude list of rat check

sundapeng:
Fixing: https://issues.apache.org/jira/browse/CRYPTO-25

@winningsix


winningsix:
Thanks @sundapeng for finding this. We could add license as annotation for source file.

sundapeng:
Thank @winningsix for your review, I think the root cause is we didn't use the same version of rat with commons. Have rebased the patch.

winningsix:
+1

sundapeng:
Committed to master. Thank @winningsix

PR #11, opened 2016-04-25
CRYPTO-9: Rename configuration name in Chimera to Apache name space

winningsix:
We need to rename configuration names under the Apache name space.

### DIFF
--- a/src/main/java/org/apache/commons/crypto/conf/ConfigurationKeys.java
+++ b/src/main/java/org/apache/commons/crypto/conf/ConfigurationKeys.java
@@ -25,25 +25,19 @@ import org.apache.commons.crypto.cipher.JceCipher;
 public class ConfigurationKeys {
 
   /**
-   * The prefix named with project name.
+   * The prefix of crypto configuration.
    */
-  public static final String CHIMERA_PREFIX = "chimera.";
+  public static final String CONF_PREFIX = "commons.crypto.";
 
   /**
    * The filename of configuration file.
    */
-  public static final String CHIMERA_SYSTEM_PROPERTIES_FILE =
-      CHIMERA_PREFIX + "properties";
-
-  /**
-   * The prefix of crypto configuration.
-   */
-  public static final String CONF_PREFIX =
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/commons/crypto/utils/Utils.java
```diff
@@ -79,15 +79,15 @@ private static void loadChimeraSystemProperties() {
       Enumeration<?> names = props.propertyNames();
       while (names.hasMoreElements()) {
         String name = (String) names.nextElement();
-        if (name.startsWith("chimera.")) {
+        if (name.startsWith("commons.crypto.")) {
```
jerrychenhf: How about use CONF_PREFIX constant instead of string here?

src/main/java/org/apache/commons/crypto/conf/ConfigurationKeys.java
```diff
@@ -25,25 +25,19 @@
 public class ConfigurationKeys {
 
   /**
-   * The prefix named with project name.
+   * The prefix of crypto configuration.
    */
-  public static final String CHIMERA_PREFIX = "chimera.";
+  public static final String CONF_PREFIX = "commons.crypto.";
```
jerrychenhf: shall we use "commons.crypto"?

src/main/java/org/apache/commons/crypto/utils/Utils.java
```diff
@@ -57,17 +57,17 @@
   private static final int AES_BLOCK_SIZE = AES_CTR_NOPADDING.getAlgorithmBlockSize();
 
   static {
-    loadChimeraSystemProperties();
+    loadCommonsCryptoSystemProperties();
```
jerrychenhf: Suggest to remove the string like "CommonsCrypto" from the method name for simplicity.

winningsix:
Thanks @jerrychenhf for your review. PR is updated.

jerrychenhf:
+1

winningsix:
Thanks @jerrychenhf for your review. I have committed to the upstream.

PR #17, opened 2016-04-27
Change the value of DECRYPT_MODE to 2 to be consistent with JDK

dianfu:
This PR is to fix CRYPTO-29

### DIFF
--- a/src/main/java/org/apache/commons/crypto/cipher/Cipher.java
+++ b/src/main/java/org/apache/commons/crypto/cipher/Cipher.java
@@ -42,7 +42,7 @@ public interface Cipher extends Closeable {
    * A constant representing decrypt mode.  The mode constant to be used 
    * when calling init method of the Cipher.
    */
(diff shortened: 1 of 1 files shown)


kexianda:
@dianfu , 'final ' seems better.
```java
public static final int ENCRYPT_MODE = 1;
public static final int DECRYPT_MODE = 2;
```

dianfu:
@kexianda Thanks for the review. Make sense. Updated the PR.

kexianda:
@dianfu 
oops, these two variable are in the interface.' public static final' seems reduntant.
Variables declared in Interface are implicitly public static final

jerrychenhf:
+1 
@dianfu 
Please also comment in the JIRA that using the constant instead of enum is consistent with JCE.

dianfu:
@jerrychenhf  Thanks for the review. Have commented in the JIRA about this.

PR #19, opened 2016-04-27
CRYPTO-30: Mutable fields should be private

winningsix:
Mutable fields should be private or possibly package-protected, never protected or public.

### DIFF
--- a/src/main/java/org/apache/commons/crypto/stream/CipherInputStream.java
+++ b/src/main/java/org/apache/commons/crypto/stream/CipherInputStream.java
@@ -50,25 +50,25 @@ public class CipherInputStream extends InputStream implements
   private final byte[] oneByteBuf = new byte[1];
 
   /**The Cipher instance.*/
-  protected final Cipher cipher;
+  final Cipher cipher;
 
   /**The buffer size.*/
-  protected final int bufferSize;
+  final int bufferSize;
 
   /**Crypto key for the cipher.*/
-  protected final Key key;
+  final Key key;
 
   /** the algorithm parameters */
-  protected final AlgorithmParameterSpec params;
+  final AlgorithmParameterSpec params;
 
   /** Flag to mark whether the input stream is closed.*/
-  protected boolean closed;
+  private boolean closed;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/commons/crypto/stream/CipherInputStream.java
```diff
@@ -430,6 +429,42 @@ protected Cipher getCipher() {
...
+  public byte[] getIv() {
+    return iv;
+  }
+
+  /**
+   * Set the IV.
+   *
+   * @param iv the IV.
+   */
+  public void setIv(byte[] iv) {
```
jerrychenhf: Do we really need setIv method? it is passed in constructor and we may don't need to change it.

src/main/java/org/apache/commons/crypto/stream/CipherInputStream.java
```diff
@@ -430,6 +429,42 @@ protected Cipher getCipher() {
...
+  public Input getInput() {
+    return input;
+  }
+
+  /**
+   * Set the input.
+   *
+   * @param input the input
+   */
+  public void setInput(Input input) {
```
jerrychenhf: Do we need public setInput? the same as setIv.

src/main/java/org/apache/commons/crypto/stream/CipherOutputStream.java
```diff
@@ -388,4 +386,112 @@ protected void freeBuffers() {
     Utils.freeDirectBuffer(inBuffer);
     Utils.freeDirectBuffer(outBuffer);
   }
+
```
jerrychenhf: We may need to consider whether we need a setXXX method to expose it to external change. If not allowed we should not expose the method.

jerrychenhf:
Xianda has a bigger change to refine the parameters. So you keep work after Xianda's code merge or for time, we can work based on his code to avoid conflicts.

jerrychenhf:
+1

winningsix:
Thanks @jerrychenhf for your review. Committed to the upstream at 98bec4c

PR #22, opened 2016-04-28
CRYPTO-31: Make fields final wherever possible

winningsix:
Change the following fields as final:
```
JavaSecureRandom.instance
OpensslSecureRandom.fallback
OpensslSecureRandom.nativeEnabled
ChannelInput.channel
StreamInput.buf
StreamOutput.bufferSize
StreamOutput.buf
StreamOutput.out
static NativeCodeLoader.nativeCodeLoaded
static OSInfo.archMapping
static ReflectionUtils.classLoader
```


src/main/java/org/apache/commons/crypto/stream/input/StreamInput.java
```diff
@@ -28,7 +28,7 @@
 public class StreamInput implements Input {
   private byte[] buf;
   private int bufferSize;
```
jerrychenhf: Can the bufferSize be final?

jerrychenhf:
LGTM

PR #41, opened 2016-05-06
CRYPTO-56: add informative error message, keep consistent with JDK

kexianda:
[CRYPTO-56](https://issues.apache.org/jira/browse/CRYPTO-56) 
JDK's error message is more informative. it would be better to keep consistent with JDK


sundapeng:
LGTM, thank @kexianda

sundapeng:
Thank @kexianda , committed to master

PR #42, opened 2016-05-09
CRYPTO-57: Fix build on Mac OS

britter:
This change adds the location of OpenSSL 101 to the Compiler and Linker parameters. After this change it is no longer necessary to call brew link --force openssl101, which may break other builds.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -215,8 +215,8 @@ Linux-armhf_COMMONS_CRYPTO_FLAGS:=
 Mac-x86_CC        := gcc -arch i386
 Mac-x86_CXX       := g++ -arch i386
 Mac-x86_STRIP     := strip -x
-Mac-x86_CFLAGS    := -Ilib/inc_mac -I$(JAVA_HOME)/include -O2 -fPIC -mmacosx-version-min=10.4 -fvisibility=hidden -I/usr/local/opt/openssl101/include
-Mac-x86_CXXFLAGS  := -Ilib/inc_mac -I$(JAVA_HOME)/include -O2 -fPIC -mmacosx-version-min=10.4 -fvisibility=hidden -I/usr/local/opt/openssl101/include
+Mac-x86_CFLAGS    := -Ilib/inc_mac -I$(JAVA_HOME)/include -O2 -fPIC -mmacosx-version-min=10.4 -fvisibility=hidden -I/usr/local/include
+Mac-x86_CXXFLAGS  := -Ilib/inc_mac -I$(JAVA_HOME)/include -O2 -fPIC -mmacosx-version-min=10.4 -fvisibility=hidden -I/usr/local/include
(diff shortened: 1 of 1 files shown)


britter:
I've just realized, that you still have to call `brew link --force openssl101`. Otherwise all tests will fail.

sundapeng:
Thank @britter for your fix. the fix sounds good to me.
Could you also update BUILDING.txt, so that I can try it on my local environment.

britter:
@sundapeng I think there is no change to BUILDING.txt necessary, since I discovered that `brew link --force` is still required. For some reason the build fails on my machine without the change to Makefile.common. I don't know, why it is working on your machine...

I've modified the commit message.

sundapeng:
Hi @britter , Could help to check if `openssl` is existed at `/usr/local/include`

sundapeng:
Thank @britter , I will commit the fix tomorrow.

britter:
@sundapeng sorry, I have been on vacation last week. I'll look through your comments today and update the PR if necessary. Thank you!

sundapeng:
Thank @britter , please feel free to let me if you have any question.

britter:
Hello @sundapeng 

> Hi @britter , Could help to check if openssl is existed at /usr/local/include

```
$ ll /usr/local/include/ | grep openssl
lrwxr-xr-x   1 bene  admin    43  9 Mai 21:36 openssl -> ../Cellar/openssl101/1.0.1t/include/openssl
```

sundapeng:
Thank @britter , in my environment, it seems compiler can found `/usr/local/include/` automatically. 
Could you help to check if it is work after changed `/usr/local/opt/openssl101/include` to ` /usr/local/include/` or `/usr/local/include/openssl`. Because it may be not good to add a directory with version like `/usr/local/opt/openssl101/include`

britter:
@sundapeng The build works on my machine using the script provided by @salyh in the JIRA issue.

sundapeng:
Thank @britter , may I know which works on your machine? `/usr/local/opt/openssl101/include`, `/usr/local/include/` or `/usr/local/include/openssl`, do you need to update the pr?

britter:
Hello @sundapeng it works only with `/usr/local/include` or `/usr/local/opt/openssl101/include`. It does not work with `/usr/local/include/openssl`.

The `LINKFLAGS` still have the openssl101 path. Do you have an idea how to get rid of that as well?

Thank you!
Benedikt

sundapeng:
It's great! Thank you very much for your help.
I think `LINKFLAGS` could use `-L/usr/local/lib`. Could you help to try it again?

britter:
@sundapeng it works! PR updated.

sundapeng:
+1 . Thank @britter for your contirbution!

sundapeng:
Committed to master, Thank @britter !

PR #43, opened 2016-05-10
CRYPTO-58: Generate VERSION file using maven resource filtering.

britter:
Generating this file using maven resource filtering is less error prone.
Currenlty crypto uses a mix of sources for determining the version
number:
- NativeClassLoader first reads from META-INF and then from VERSION
- Makefile.common only uses the VERSION file

This change makes sure the version number will be set from only one
source, which is maven.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -18,7 +18,7 @@
 TARGET:=target
 SRC:=src/main/java
 SRC_NATIVE:=src/main/native
-include $(SRC)/org/apache/commons/crypto/VERSION
+include $(TARGET)/classes/org/apache/commons/crypto/VERSION
 
 ifndef JAVA_HOME
 $(error Set JAVA_HOME environment variable)
--- a/pom.xml
+++ b/pom.xml
@@ -287,6 +287,12 @@ Features
   </licenses>
 
   <build>
+    <resources>
+      <resource>
+        <directory>src/main/resources/</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
     <pluginManagement>
       <plugins>
         <plugin>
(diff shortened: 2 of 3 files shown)


sundapeng:
+1

src/main/java/org/apache/commons/crypto/utils/NativeCodeLoader.java
```diff
@@ -231,7 +231,7 @@ public static String getVersion() {
                 .getResource("/META-INF/maven/org.apache.commons.crypto/commons-crypto/pom.properties");
         if (versionFile == null) {
             versionFile = NativeCodeLoader.class
-                    .getResource("/org/apache/commons/crypto/VERSION");
```
sundapeng: Do we need this change? It seems  {{.getResource("/org/apache/commons/crypto/VERSION");}} is right.

sundapeng:
Hi @britter , I have only one comment, others are LGTM

sundapeng:
Hi @britter , I will committed it with the fix tomorrow, please feel free to comment or open a new ticket to revert it.

britter:
@sundapeng sorry, I have been on vacation last week. I'll look through your comments today and update the PR if necessary. Thank you!

sundapeng:
Thank @britter , please feel free to let me if you have any question.

src/main/java/org/apache/commons/crypto/utils/NativeCodeLoader.java
britter: I've moved the VERSION file, so it is in the same package as the `NativeClassLoader` class. This way I don't need to declare the full path. Using the full path is basically the same in this case. If you like the full path for clarity, I can change this again.

sundapeng: Thank @britter . It seems the `VERSION` file is still under `org/apache/commons/crypto/`(resources/org/apache/commons/crypto/VERSION), so I think we should use `/org/apache/commons/crypto/VERSION`, I also tested with disable ` .getResource("/META-INF/maven/org.apache.commons.crypto/commons-crypto/pom.properties");`, did I miss anything?

britter:
@sundapeng I've changed the path to `/org/apache/commons/crypto/VERSION` again for loading the version file.

sundapeng:
Committed to master, thank @britter for your contribution

PR #44, opened 2016-05-25
CRYPTO-60 OpensslCipher support GCM

kexianda:
The interface is similar to JCE. In encryption mode, the authenticated tag information is appended at the end of output.
In decryption mode, the authenticated tag should appended at the end of input. If the encrypted data is tampered, it will throw an AEADBadTagException.

### DIFF
--- a/src/main/java/org/apache/commons/crypto/cipher/CryptoCipherFactory.java
+++ b/src/main/java/org/apache/commons/crypto/cipher/CryptoCipherFactory.java
@@ -21,6 +21,7 @@ import java.security.GeneralSecurityException;
 import java.util.List;
 import java.util.Properties;
 
+import org.apache.commons.crypto.conf.ConfigurationKeys;
 import org.apache.commons.crypto.utils.ReflectionUtils;
 import org.apache.commons.crypto.utils.Utils;
 
@@ -47,8 +48,7 @@ public class CryptoCipherFactory {
     public static CryptoCipher getInstance(String transformation,
             Properties props) throws GeneralSecurityException {
 
-        List<String> klasses =  Utils.splitClassNames(
-                Utils.getCipherClassString(props), ",");
+        List<String> klasses =  Utils.splitClassNames(getCipherClassString(props), ",");
         CryptoCipher cipher = null;
 
         StringBuilder errorMessage = new StringBuilder("CryptoCipher ");
@@ -93,6 +93,25 @@ public class CryptoCipherFactory {
     public static CryptoCipher getInstance(String transformation)
             throws GeneralSecurityException {
         return getInstance(transformation, new Properties());
+    }
+
+    /**
+     * Gets the cipher class.
+     *
+     * @param props The <code>Properties</code> class represents a set of
+     *        properties.
+     * @return the cipher class based on the props.
+     */
+    private static String getCipherClassString(Properties props) {
+        final String configName = ConfigurationKeys.CIPHER_CLASSES_KEY;
+        String cipherClassString = props.getProperty(configName) != null ? props
+                .getProperty(configName, ConfigurationKeys.CIPHER_CLASSES_DEFAULT)
+                : System.getProperty(configName,
+                ConfigurationKeys.CIPHER_CLASSES_DEFAULT);
+        if (cipherClassString.isEmpty()) {
+            cipherClassString = ConfigurationKeys.CIPHER_CLASSES_DEFAULT;
+        }
+        return cipherClassString;
     }
 
 }
--- a/src/main/java/org/apache/commons/crypto/utils/Utils.java
+++ b/src/main/java/org/apache/commons/crypto/utils/Utils.java
@@ -137,25 +137,6 @@ public final class Utils {
     }
 
     /**
-     * Gets the cipher class.
-     *
-     * @param props The <code>Properties</code> class represents a set of
-     *        properties.
-     * @return the cipher class based on the props.
-     */
-    public static String getCipherClassString(Properties props) {
-        final String configName = ConfigurationKeys.CIPHER_CLASSES_KEY;
-        String cipherClassString = props.getProperty(configName) != null ? props
-                .getProperty(configName, ConfigurationKeys.CIPHER_CLASSES_DEFAULT)
-                : System.getProperty(configName,
-                ConfigurationKeys.CIPHER_CLASSES_DEFAULT);
-        if (cipherClassString.isEmpty()) {
-            cipherClassString = ConfigurationKeys.CIPHER_CLASSES_DEFAULT;
-        }
-        return cipherClassString;
-    }
-
-    /**
      * Gets path of native library.
      *
      * @return the path of native library.


src/main/java/org/apache/commons/crypto/cipher/CipherTransformation.java
```diff
@@ -28,7 +28,9 @@
     /** A crypto transformation representing AES/CBC/NoPadding */
     AES_CBC_NOPADDING("AES/CBC/NoPadding", 16),
     /** A crypto transformation representing AES/CBC/PKCS5Padding */
-    AES_CBC_PKCS5PADDING("AES/CBC/PKCS5Padding", 16);
+    AES_CBC_PKCS5PADDING("AES/CBC/PKCS5Padding", 16),
+
```
sundapeng: Better to add javadoc here.

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
```diff
@@ -323,11 +459,56 @@ public void clean() {
         }
     }
 
+    public void updateAAD(byte[] aad) {
```
sundapeng: Please add @javadoc

src/main/java/org/apache/commons/crypto/cipher/JceCipher.java
```diff
@@ -197,6 +197,11 @@ public int doFinal(byte[] input, int inputOffset, int inputLen,
                 outputOffset);
     }
 
+    @Override
+    public void updateAAD(byte[] aad) {
```
sundapeng: Javadoc

src/main/java/org/apache/commons/crypto/cipher/CryptoCipher.java
```diff
@@ -156,4 +156,32 @@ int doFinal(ByteBuffer inBuffer, ByteBuffer outBuffer)
...
+     * operating in either GCM or CCM mode and one of the {@code update}
+     * methods has already been called for the active
+     * encryption/decryption operation
+     * @throws UnsupportedOperationException if the corresponding method
+     * in the {@code CipherSpi} has not been overridden by an
+     * implementation
+     *
+     * @since
+     */
+    void updateAAD(byte[] aad)
```
sundapeng: Do we need `updateAAD(ByteBuffer byteBuffer )`?

src/main/java/org/apache/commons/crypto/cipher/CryptoCipher.java
```diff
@@ -156,4 +156,31 @@ int doFinal(ByteBuffer inBuffer, ByteBuffer outBuffer)
...
+     * methods has already been called for the active
+     * encryption/decryption operation
+     * @throws UnsupportedOperationException if the corresponding method
+     * has not been overridden by an implementation
+     *
+     * @since
+     */
+    void updateAAD(byte[] aad)
+            throws IllegalArgumentException, IllegalStateException, UnsupportedOperationException;
 }
```
sundapeng: Do we need `updateAAD(ByteBuffer byteBuffer )`?

src/main/java/org/apache/commons/crypto/cipher/CryptoCipher.java
kexianda: yes, JceCipher should also support GCM mode. 
Hmm.. it would be better to add some test cases for JceCipher

src/main/java/org/apache/commons/crypto/cipher/CryptoCipher.java
kexianda: thanks. fixed.

kexianda:
there is a bug in current implementation.  GCM spec mentioned that only return recovered data after tag is successfully verified.  
Hold on. I will update the PR soon.

kexianda:
PR is updated: 1. fix a bug( return recovered data after tag is verified) 2. add some test cases

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
```diff
@@ -43,9 +49,20 @@
...
+    static final int DEFAULT_TAG_LEN = 16;
+
+    private int mode = Openssl.DECRYPT_MODE;
+
+    // buffer for AAD data; if consumed, set as null
+    private byte[] aadBuffer;
+    private int tagBitLen = -1;
+
+    // buffer for storing input in decryption, not used for encryption
+    private ByteArrayOutputStream ibuffer = null;
```
jerrychenhf: Is there any more efficient way instead of buffering all the input data in byte array?
Java document said this:
To avoid implementations having to internally buffer ciphertext, all AAD data must be supplied to GCM/CCM implementations (via the updateAAD methods) before the ciphertext is processed (via the update and doFinal methods).

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
```diff
@@ -323,11 +622,42 @@ public void clean() {
...
+     * Calls to this method provide AAD to the cipher when operating in
+     * modes such as AEAD (GCM).  If this cipher is operating in
+     * either GCM mode, all AAD must be supplied before beginning
+     * operations on the ciphertext (via the {@code update} and
+     * {@code doFinal} methods).
+     *
+     * @param aad the buffer containing the Additional Authentication Data
+     *
+     */
+    public void updateAAD(byte[] aad) {
```
jerrychenhf: Is it right that updateAAD simply overwrite the aadbuffer?

I thought it means append more aad data.

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
```diff
@@ -323,11 +622,42 @@ public void clean() {
...
+        // must be called after initialized.
+        aadBuffer = aad.clone();
+    }
+
     /** Checks whether context is initialized. */
     private void checkState() {
         Utils.checkState(context != 0);
     }
 
+    private void processAAD() {
```
jerrychenhf: processAAD is little strange. It was called in every call of update or doFinal.

Does Java do in the similar way? Shall this be done when updateAAD is called?

jerrychenhf:
My major concern is the processing of the decryption with buffers.  

We may also be able to seek the possibility the minimize the code different between GCM and other modes in the code block shared by all modes.

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
kexianda: According to GCM Spec, after the AAD is processed, GCM cipher got a initial intermediate tag value (denoted as X in the Spec).  Then, X is XORed with the ciphertext blocks one by one.
From my understanding, Java doc means that, if the initial intermediate tag value is NOT ready( AAD is not processed) before encryption(update/doFinal), to get the intermediate tag value, GCM cipher has to internally buffer ciphertext for later calculation of  intermediate tag value.  If AAD data is ready, GCM cipher only has to buffer the intermediate tag value.

In a word, gcm implementation has to avoid buffering ciphertext not the AAD.

Thanks for your reminder, I find a bug. we can not use byte[] as buffer, because updateAAD()  can be called zero or more times. the aad data should be concatenated. it seems that ByteArrayOutputStream is suitable for buffering aad. I will update soon.
```java
//updateAAD()  can be called zero or more times
cipher.updateAAD(aad1);
cipher.updateAAD(aad2);
```

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
kexianda: Thanks, you are right. I will update it soon.

src/main/java/org/apache/commons/crypto/cipher/Openssl.java
kexianda: Yes, we have to call processAAD() before any update/doFinal.  To calculate the intermediate tag value, AAD should be processed before update/doFinal

PR #47, opened 2016-05-25
Introduce complete JNA implementation of cipher and cryptorandom

salyh:
Before merging this we should benchmark it and compare performance to JNI openssl binding

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,7 @@ Features
     <junit.version>4.8.2</junit.version>
     <commons-logging.version>1.1.3</commons-logging.version>
     <slf4j-api.version>1.7.10</slf4j-api.version>
+    <jna.version>4.2.2</jna.version>
   </properties>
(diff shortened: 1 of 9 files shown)


sundapeng:
Thank @salyh for your great work.
I prefer the JNA implementation if performance does not decrease too much. I will do some performance test on it.

sebbASF:
The implementation works OK for many cases, but the Cipher class crashes the JVM if
- init() is called after close()
- close() is called twice

Commenting out the call to OpensslNativeJna.EVP_CIPHER_CTX_free(context); in close() fixes the problem, but that means the context is never freed.

salyh:
this seems to be merged somehow with https://github.com/apache/commons-crypto/commit/b6006e563f63c745c964a87fae459200ae24573b

sundapeng:
Thank @salyh for your contribution.

PR #52, opened 2016-06-15
CRYPTO-71 Hide the implementation of OpensslCipher

kexianda:
classes Openssl and OpensslNative are implemetation, they should be package private


src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -313,7 +313,6 @@ protected CryptoCipher getCipher(String cipherClass) throws IOException {
 
     @Test
     public void testReadWrite() throws Exception {
-        Assert.assertEquals(null, Openssl.getLoadingFailureReason());
```
winningsix: Is it possible to make it a method of Openssl member method instead of removing the check here?

kexianda: @winningsix ,  if the underlying engine failed,  OpensslCipher will throw an exception with the loadingFailureReason message. if the native is not loaded, the UT will also failed (an exception is thrown out).
It seems that there is no need to add an interface, which is exposed to client user, to check if the native is loaded successfully. An exception seems enough and friendly for the client user.

winningsix: If native code is failing to load, the exception will be caught by throwable block. This is designed for fall back mechanism for cipher loader.

winningsix:
LGTM +1

sundapeng:
Committed to master. Thank @kexianda and @winningsix

PR #57, opened 2016-06-21
CRYPTO-86 remove the Properties field in CryptoCipher

kexianda:
[CRYPTO-86](https://issues.apache.org/jira/browse/CRYPTO-86)


src/main/java/org/apache/commons/crypto/stream/PositionedCryptoInputStream.java
```diff
@@ -55,6 +55,8 @@
      */
     private final Queue<CipherState> cipherPool = new ConcurrentLinkedQueue<CipherState>();
 
+    private final Properties props;
```
sundapeng: Better to add comments here.

sundapeng:
Thank @kexianda ,only one comment left, other parts are LGTM.

sundapeng:
+1

PR #58, opened 2016-06-21
CRYPTO-74: Full class names make code more difficult to update

sundapeng:
Initial patch for https://issues.apache.org/jira/browse/CRYPTO-74


sundapeng:
Merged!

PR #59, opened 2016-06-21
CRYPTO-87: Reduce util package API footprint

britter:
See https://issues.apache.org/jira/browse/CRYPTO-87


sundapeng:
Thank @britter , +1, it seems we should rebase the pr.

britter:
Branch rebased!

PR #61, opened 2016-06-23
CRYPTO-82: CipherTransformation is an enum which limits the possible …

kexianda:
[CRYPTO-82](https://issues.apache.org/jira/browse/CRYPTO-82) CipherTransformation is an enum which limits the possible transformations

delete enum getTransformation, use String type transformation.
And add two methods getAlgorithm() / getBlockSize for CryptoCipher,
Keep the API similar to JCE


sundapeng:
@kexianda I think we should rebase the pr.

kexianda:
@sundapeng . rebased now, please help review.

sundapeng:
Thank @kexianda ,committed to master.

PR #62, opened 2016-06-23
CRYPTO-89: more robust native code to eliminate memory leak

kexianda:
- 1. error handling in OpensslNative is hard to read and maintain

C doesn't support exceptions, try-finally blocks, or RAII at all. A typical approach is to separate releasing of resources at the end of the function and jump there with gotos in the case of error. This way the cleanup code need not be duplicated.

- 2. memory leak cases
(a). resource is allocated, but not de-allocated when error occurs
(b). openssl evp context pointer could be overwritten as null, but the context is not de-allocated.

### DIFF
--- a/src/main/native/org/apache/commons/crypto/cipher/OpensslNative.c
+++ b/src/main/native/org/apache/commons/crypto/cipher/OpensslNative.c
@@ -247,6 +247,11 @@ JNIEXPORT jlong JNICALL Java_org_apache_commons_crypto_cipher_OpensslNative_init
     (JNIEnv *env, jclass clazz, jlong ctx, jint mode, jint alg, jint padding,
     jbyteArray key, jbyteArray iv)
 {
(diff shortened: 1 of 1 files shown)


kexianda:
hold on. i will update it

sundapeng:
Thank @kexianda for your contribution, please let me know if the pr is ready.

kexianda:
updated. fix four issues: @sundapeng, please help review
- 1. using goto so that the cleanup code need not be duplicated
- 2. make sure that ByteArrayElements is released
- 3. add error handling for EVP_CIPHER_CTX_set_padding
- 4. if openssl context is ready,  Don't free it in OpensslNative_init() even if error occur, because the JNI throw exceptions and java layer won't set context pointer as NULL. so, the context can be freed again if Openssl.finalize() is called by GC. it may cause JVM crash

PR #63, opened 2016-06-29
CRYPTO-93: add a factory method to provide simpler API

kexianda:
[CRYPTO-93](https://issues.apache.org/jira/browse/CRYPTO-93)
 use the better hardware random by default

### DIFF
--- a/src/main/java/org/apache/commons/crypto/examples/RandomExample.java
+++ b/src/main/java/org/apache/commons/crypto/examples/RandomExample.java
@@ -32,13 +32,36 @@ import org.apache.commons.crypto.random.CryptoRandomFactory;
 public class RandomExample {
 
     public static void main(String []args) throws GeneralSecurityException, IOException {
+        exampleWithdefaultHardwareRandom();
+        exampleWithUserDefinedRandom();
+    }
+
+    public static void exampleWithdefaultHardwareRandom()
+            throws GeneralSecurityException, IOException {
         //Constructs a byte array to store random data.
         byte[] key = new byte[16];
-        byte[] iv = new byte[32];
+
+        //Gets the 'CryptoRandom' instance.
+        CryptoRandom random = CryptoRandomFactory.getCryptoRandom();
+
+        // Generate random bytes and places them into the byte arrays.
+        random.nextBytes(key);
+
+        //Closes the CryptoRandom.
+        random.close();
+
+        // Show the generated output
+        System.out.println(Arrays.toString(key));
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/crypto/examples/RandomExample.java
```diff
@@ -32,29 +32,62 @@
...
+
+            //Closes the CryptoRandom.
+            random.close();
+
+            // Show the generated output
+            System.out.println(Arrays.toString(key));
+
+        } catch (Exception ex) {
+            ex.printStackTrace();
+        }
```
sebbASF: This is sample code so should be as simple as possible.
Drop the try/catch and throw Exception from the method instead

src/main/java/org/apache/commons/crypto/examples/RandomExample.java
```diff
@@ -32,29 +32,62 @@
...
+            System.out.println(Arrays.toString(key));
+            System.out.println(Arrays.toString(iv));
 
-        // Show the generated output
-        System.out.println(Arrays.toString(key));
-        System.out.println(Arrays.toString(iv));
+        } catch (Exception ex) {
+            ex.printStackTrace();
+        }
     }
```
sebbASF: Ditto; don't try/catch - throw instead

src/main/java/org/apache/commons/crypto/random/CryptoRandomFactory.java
```diff
@@ -134,6 +134,21 @@ public static CryptoRandom getCryptoRandom(Properties props)
     }
 
     /**
+     * Get a OpenSSL CryptoRandom by default.
+     * If OpenSSL is unavailable qand fallback is enable, we will use Java Random
```
sebbASF: Fallback no longer exists

src/main/java/org/apache/commons/crypto/examples/RandomExample.java
```diff
@@ -32,29 +32,62 @@
...
-        byte[] iv = new byte[32];
+        exampleWithdefaultHardwareRandom();
+        exampleWithUserDefinedRandom();
+    }
+
+    public static void exampleWithdefaultHardwareRandom() {
+        try {
+            //Constructs a byte array to store random data.
+            byte[] key = new byte[16];
+
```
sebbASF: This does not agree with the other example as it only uses a single byte array

kexianda:
thanks @sebbASF for your comments. make sense. updated.

PR #64, opened 2016-06-29
CRYPTO-92: Handling default properties; allow SystemProperties to be ignored

sundapeng:
https://issues.apache.org/jira/browse/CRYPTO-92


sebbASF:
This is not quite what I meant.
The user should be able to control which defaults are used, not have it decided for them.
If the user wants to default to a different set of properties, then they should provide that as part of the Properties instance.
The methods should only use the props parameter they are given.
If the user wants to add defaults, they can call Utils.getProperties(props) themselves.
This would also be more efficient than creating new Properties in each method.

PR #66, opened 2016-07-04
CRYPTO-96: OpenSSL Random implementation silently falls back to Java

sundapeng:
https://issues.apache.org/jira/browse/CRYPTO-96

### DIFF
--- a/src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
+++ b/src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
@@ -17,7 +17,7 @@
  */
 package org.apache.commons.crypto.random;
 
-import java.security.NoSuchAlgorithmException;
+import java.security.GeneralSecurityException;
 import java.util.Properties;
 import java.util.Random;
 
@@ -44,8 +44,6 @@ import org.apache.commons.crypto.utils.Utils;
 class OpenSslCryptoRandom extends Random implements CryptoRandom {
     private static final long serialVersionUID = -7828193502768789584L;
 
-    /** If native SecureRandom unavailable, use java SecureRandom */
-    private final JavaCryptoRandom fallback;
     private static final boolean nativeEnabled;
 
     static {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
```diff
@@ -44,46 +43,25 @@
...
     // N.B. this class is not public/protected so does not appear in the main Javadoc
     // Please ensure that property use is documented in the enum CryptoRandomFactory.RandomProvider
-    public OpenSslCryptoRandom(Properties props)
-            throws NoSuchAlgorithmException {
-        //fallback needs to be initialized here in any case cause even if
-        //nativeEnabled is true OpenSslCryptoRandomNative.nextRandBytes may fail
-        fallback = new JavaCryptoRandom(props);
+    public OpenSslCryptoRandom(Properties props) {
+        // Generates one byte to verify Openssl random
+        next(1);
```
sebbASF: Surely one needs to check the return value here?

src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
```diff
@@ -93,9 +71,7 @@ public OpenSslCryptoRandom(Properties props)
      */
     @Override
     public void nextBytes(byte[] bytes) {
-        if (!nativeEnabled || !OpenSslCryptoRandomNative.nextRandBytes(bytes)) {
-            fallback.nextBytes(bytes);
-        }
+        OpenSslCryptoRandomNative.nextRandBytes(bytes);
```
sebbASF: Again, the return value needs to be checked

src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
```diff
@@ -140,8 +116,5 @@ final protected int next(int numBits) {
      */
```
sebbASF: Javadoc needs adjusting now that the code has changed

src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
```diff
@@ -44,46 +43,25 @@
...
-    private static final boolean nativeEnabled;
-
     static {
-        boolean opensslLoaded = false;
         if (Crypto.isNativeCodeLoaded()) {
             try {
                 OpenSslCryptoRandomNative.initSR();
-                opensslLoaded = true;
             } catch (Exception t) {// NOPMD
             }
```
sebbASF: What if the native code is not loaded, or initSR fails?
This should surely be noted here in a final boolean and used in the ctor and methods?

src/main/java/org/apache/commons/crypto/random/OpenSslCryptoRandom.java
```diff
@@ -44,46 +43,25 @@
...
-    }
-
-    /**
-     * Judges whether loading native library successfully.
-     *
-     * @return true if loading library successfully.
-     */
-    public static boolean isNativeCodeLoaded() {
-        return nativeEnabled;
     }
```
sebbASF: Surely this method is still needed?

sundapeng:
Thank @sebbASF ，patch updated

sebbASF:
Thanks!

I've been looking into this further, and I think we could/should provide more specific feedback for failures.
For example the static block could save the exception when initSR fails.
Also nextBytes does not need to check nativeEnabled.
See attached patch for sample implementation.
[crypto96-patch.txt](https://github.com/apache/commons-crypto/files/349504/crypto96-patch.txt)

sundapeng:
Thank @sebbASF for your comment. I will commit it.

PR #71, opened 2016-10-19
[CRYPTO-125] CryptoOutputStream does not call write in a loop when underlying channel works in non-block mode

cjjnjust:
Add a loop to drain out the internal buffer.


sundapeng:
+1

kexianda:
two concerns:
1. 
```
// if output reaches at it's end, but outBuffer still has remaining, infinite loop?
while (outBuffer.hasRemaining()) {
     output.write(outBuffer);
}
```

2. Does CryptoOutputStream#encryptFinal() has to handle this non-block case ?

sundapeng:
+1, Thank @kexianda and @cjjnjust for the contribution. I plan to commit it later, please let know if you have any concern.

kexianda:
LGTM
+1

PR #72, opened 2016-11-02
CRYPTO-127: CryptoInputStream#read should handle non-block case

kexianda:


sundapeng:
+1

src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java
```diff
@@ -255,8 +255,12 @@ public int read(byte[] b, int off, int len) throws IOException {
...
         // No data in the out buffer, try read new data and decrypt it
-        int nd = decryptMore();
-        if (nd <= 0) {
+        // we loop for new data
+        int nd = 0;
+        while (nd == 0) {
+            nd = decryptMore();
+        }
+        if (nd < 0) {
             return nd;
```
sundapeng: if we want to change to loop here, the return value should also be updated.

sundapeng:
Thank Xianda, I have only one comment.

src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java
kexianda: @sundapeng 
Hmm...,  it seems OK now. 
It's would be better if we add some testcases in future

sundapeng: Yes, I think it should be okay, I misunderstood it.

sundapeng:
+1

sundapeng:
Thank @kexianda

PR #73, opened 2016-11-14
CRYPTO-129: Change access of instance variables from package private to private (or protected if appropriate)

jianguotian:
[Crypto 129](https://issues.apache.org/jira/browse/CRYPTO-129)
Modifier of some  variables has changed.


sundapeng:
+1, thank @jianguotian

PR #74, opened 2017-03-01
[CRYPTO-135]CryptoOutputStream is always blocking

cjjnjust:
int write(ByteBuffer src) should not do blocking write but return actual write out bytes which underlying output writes. 

I add a NetworkStreamTest to test such non-blocking write.

### DIFF
--- a/src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
+++ b/src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
@@ -18,6 +18,17 @@
 
 package org.apache.commons.crypto.stream;
 
+import org.apache.commons.crypto.cipher.CryptoCipher;
+import org.apache.commons.crypto.stream.output.ChannelOutput;
+import org.apache.commons.crypto.stream.output.Output;
+import org.apache.commons.crypto.stream.output.StreamOutput;
+import org.apache.commons.crypto.utils.Utils;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.ShortBufferException;
+import javax.crypto.spec.IvParameterSpec;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
@@ -27,17 +38,6 @@ import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.spec.AlgorithmParameterSpec;
 import java.util.Properties;
-import javax.crypto.BadPaddingException;
-import javax.crypto.Cipher;
-import javax.crypto.IllegalBlockSizeException;
-import javax.crypto.ShortBufferException;
-import javax.crypto.spec.IvParameterSpec;
-
-import org.apache.commons.crypto.cipher.CryptoCipher;
-import org.apache.commons.crypto.stream.output.ChannelOutput;
-import org.apache.commons.crypto.stream.output.Output;
-import org.apache.commons.crypto.stream.output.StreamOutput;
-import org.apache.commons.crypto.utils.Utils;
 
 /**
  * {@link CryptoOutputStream} encrypts data and writes to the under layer
@@ -78,6 +78,8 @@ public class CryptoOutputStream extends OutputStream implements
      * at outBuffer.limit().
      */
     ByteBuffer outBuffer; // package protected for access by crypto classes; do not expose futher
+
+    boolean blocking;
 
     /**
      * Constructs a {@link CryptoOutputStream}.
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +298,39 @@ public boolean isOpen() {
...
-            final int space = inBuffer.remaining();
-            if (remaining < space) {
+            int position = src.position();
+            int written;
+            final int oldLimit = src.limit();
+            int space = inBuffer.remaining();
+
+            if(remaining < inBuffer.remaining()) {
                 inBuffer.put(src);
-                remaining = 0;
```
sundapeng: IMO, after the `src` put into buffer, we should break the loop directly. `encrypt` should only happen when buffer is full or doFinal is called.

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +298,39 @@ public boolean isOpen() {
...
+
+        int bytes_written = 0;
+        int remaining = src.remaining();
+
         while (remaining > 0) {
-            final int space = inBuffer.remaining();
-            if (remaining < space) {
+            int written;
+            final int oldLimit = src.limit();
+            int space = inBuffer.remaining();
```
sundapeng: I think `written`, `oldLimit` and `space` should move to scope of `else`.

cjjnjust:
@sundapeng @winningsix  Any other comments?

cjjnjust:
@sundapeng @winningsix  do we have any other consideration for this? any other comments?

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +301,47 @@ public boolean isOpen() {
...
                 src.limit(oldLimit);
-
                 remaining -= space;
-                encrypt();
+                bytesWritten += space;
+
+                encryptNonBlocking();
+                written = output.write(outBuffer);
+
+                // Underlying output maybe blocking, break out and return.
```
vanzin: s/maybe blocking/out of buffer space

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +301,47 @@ public boolean isOpen() {
...
+                if (written < space) {
+                    transferInProgress = true;
```
vanzin: Do you really need this flag? Seems to me that checking whether `output` has data left should be enough?

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -359,12 +379,32 @@ protected void encrypt() throws IOException {
...
         }
     }
 
+    /**
+     * Does the encryption in a non-blocking mode, input is {@link #inBuffer} and output is
+     * {@link #outBuffer}.
+     *
+     * @throws IOException if an I/O error occurs.
+     */
+    protected void encryptNonBlocking() throws IOException {
```
vanzin: I think instead of this method you should instead remove the following code from the existing `encrypt()`:

```
        // write to output
        while (outBuffer.hasRemaining()) {
            output.write(outBuffer);
        }
```

And move it to the existing blocking call sites (or maybe some helper method).

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +301,47 @@ public boolean isOpen() {
     @Override
     public int write(ByteBuffer src) throws IOException {
         checkStream();
-        final int len = src.remaining();
-        int remaining = len;
+
+        if (transferInProgress && outBuffer.hasRemaining()) {
+            output.write(outBuffer);
+            return 0;
```
vanzin: One improvement here would be to not return if all of `outBuffer` has been written, in which case it might still be possible to write to the output.

src/main/java/org/apache/commons/crypto/stream/CryptoOutputStream.java
```diff
@@ -298,30 +301,47 @@ public boolean isOpen() {
...
-            final int space = inBuffer.remaining();
-            if (remaining < space) {
+            int written;
+            int space = inBuffer.remaining();
+
+            if(remaining < space) {
                 inBuffer.put(src);
-                remaining = 0;
+                bytesWritten += remaining;
+                break;
```
vanzin: Hmm... here you're just buffering data without even trying to write to the output. Is that intentional? Feels like you should `encrypt()` and try to write stuff even if all the input buffer fits in the internal buffer.

But really there might be a disconnect here that I don't know is possible to fix. If the input buffer has `x` bytes, and they fit in the internal buffer, but you can only write `y` bytes (`y < x`) to the underlying channel, what happens?

If you return "I wrote x bytes", the caller may expect that those bytes will be in the final output, when that might not be the case. If you return "I wrote y bytes", it assumes that y bytes of plain text are equal to y bytes of cipher text. If you return 0, you may mess up the caller (since you've actually consumed input but are reporting 0 bytes written).

I'm starting to think that it might be better to leave this as a blocking interface, and leave it to applications to deal with non-blocking channels. e.g., thinking about Spark, you can't hook this `WritableByteChannel` implementation cleanly. It actually makes more sense to let Spark handle the data encryption, much like this code is doing internally, and do its own bookkeeping of what has been written and what has been not.

cjjnjust:
OK, please go ahead.

PR #78, opened 2018-05-14
adding support for openssl 1.1.0g in common crypto

ossdev07:
AARCH 64 architecture support was missing in the exiting common-crypto code.
After these modifcation now the common-crypto project can be used in aarch64
platform and also compatible with openssl 1.1.x.


ossdev07:
openssl version 1.1.0g is the default openssl in latest release. There are a lot changes in APIs  name in between openssl version 1.0.x and 1.1.0g.
This patch is applicable and validated with opensssl version 1.1.0g version (latest version) for AARCH64 and X86_64 architecture.

ossdev07:
this request has been verified internally with openssl 1.1.0g on amd64 as well as aarch64 platform.
please check with openssl 1.1.0g and merge it

ossdev07:
The command "eval mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V " failed. 
The JAVA_HOME environment variable is not defined correctly
This environment variable is needed to run this program
NB: JAVA_HOME should point to a JDK not a JRE

**please configured your environment so that my patch can be test and run**

vanzin:

PR #79, opened 2018-09-24
Added Support for both OpenSSL version 1.0 and 1.1

ossdev07:
1. Created .xjava file for files in which version realted calls are diffrent
2. Added code in Make.common to determine verion on running system
3. Added rules in Make file to convert from .xjava to .java and then to .class

Signed-off-by: ossdev <ossdev@puresoftware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,29 @@ endif
 NATIVE_TARGET_DIR:=$(TARGET)/classes/org/apache/commons/crypto/native/$(OS_NAME)/$(OS_ARCH)
 NATIVE_DLL:=$(NATIVE_TARGET_DIR)/$(LIBNAME)
 
-all: $(NATIVE_DLL)
+#all source files having #ifdef flag
+SOURCES_FILES := $(shell find src/main -name '*.xjava' | sort -r)
+TEST_FILES := $(shell find src/test -name '*.xjava')
+#create java file corresponding to .xjava file
+XJAVA_SRC_OBJS  := $(SOURCES_FILES:.xjava=.java)
+XJAVA_TEST_OBJS := $(TEST_FILES:.xjava=.java)
+
+#create new class file for the same
+JAVA_SRC_OBJ :=$(XJAVA_SRC_OBJS:.java=.class)
+JAVA_TEST_OBJ :=$(XJAVA_TEST_OBJS:.java=.class)
+
+%.java:%.xjava
+	@$(CPP) -D$(DEFINES) $< $@
+
+%.class : %.java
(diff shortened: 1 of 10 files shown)


ossdev07:
@vanzin will you please have a look on this PR . it is to provide support for openssl 1.1.x version along with 1.0.x ..It is verified in our setup and in travis CI for jdk 8 it is verified ..

vanzin:
I'm not a big fan of your approach. IMO the correct approach would be to detect the openssl version at runtime; all uses of openssl in this code are already done by dynamically looking for the library functions, this would be just an extension to that.

e.g. the code you have around `ENGINE_load_rdrand` in the Java side would be better in the native code, in `openssl_rand_init`, calling the appropriate openssl functions depending on what's available.

Otherwise you need different versions of this library for different versions of openssl; it's already quite a pain to build it, and that would make it worse, and you'd have to change the maven artifacts to account for the openssl version. And you'd have to do the same for other things that use this library (e.g. Spark). Lots of downsides, basically.

(And I don't really want to comment about running a pre-processor on Java source files...)

It also would be nice to check if Travis has a build image with openssl 1.1, so this code can be tested.

ossdev07:
@vanzin We can find out the openssl version at run time in c native code only but there are few APIs in java code also which has been replaced . For supporting both openssl versions we need to put both APIs on the basis of check for openssl version and we can not do in java code that's why we took this approach ...

vanzin:
The problem with your approach is not that it's doing the OpenSSL version detection in Java vs. C. The problem is that it's doing it at compile-time vs. runtime. And that means you need different builds of the library for different versions of OpenSSL. And that makes deploying the commons-crypto library very awkward.

You could do that if you somehow had both versions of the library in the same jar file, and detected the native openssl version somehow when loading the library from java. But that means doubling the amount of things you have to build to ship a new commons-crypto version.

ossdev07:
@vanzin in java code we can not detect openssl version at run time so this is the only solution we found.. Please suggest if we can achieve same by any other approach

vanzin:
There's a ton of different ways. e.g. make the JNI API independent of the OpenSSL version and do everything in C. Or expose the OpenSSL version using the JNI library and handle that in Java.

Just don't do it using compile time flags or using a pre-processor to generate Java source files.

ossdev07:
@vanzin i am checking it with some other option to provide support for both openssl version. i just want to know when maintainers are planning to update commons-crypto with openssl 1.1.0

ossdev07:
@vanzin as per your suggestion i took a new approach to support both OpenSSL (1.1 and 1.0) without using preprocessing in java code please have a look on below request 
https://github.com/apache/commons-crypto/pull/90

vanzin:

PR #80, opened 2018-09-26
CRYPTO-141: Avoid calling operations on an invalid context.

vanzin:
There are two main changes here.

The first one wraps the OpenSSL cipher context with some extra state
needed by the JNI wrapper. This allows checking whether it's safe to
call operations such as EVP_CipherUpdate, which can cause problems if
the context is not properly initialized. This more closely follows the
semantics of the Java interface.

The second change avoids cleaning up the OpenSSL context when an error
happens. Doing that would invalidate the memory used by the context,
but the Java code would still have the address of the now invalid
context, and calling another JNI method would crash the JVM. This
ties the lifecycle of the OpenSSL context to the Java cipher instance.

As part of those changes, all uses of EVP_CIPHER_CTX_cleanup were
removed; the remaining cleanup paths use EVP_CIPHER_CTX_free,
which is its replacement.

There are a few more changes that are not required for the above but
helped with testing and debugging:

- some very minor changes in Java code, especially using final fields
  where they make sense.
- enable -Wall and -Werror when compiling on linux64.
- fixed some minor issues with error handling in the native code.


vanzin:
@winningsix are you still working with this code? Would be good to get a second set of eyes on this.

vanzin:
@squito just in case you don't have enough to do and want to review some more stuff.

squito:
sorry @vanzin I think this would be a really hard change for me to review.  hopefully you can find somebody more familiar with this

PR #90, opened 2018-10-30
commons-crypto OpenSSL's to support  both 1.0 and 1.1

ossdev07:
1. Have added new files with changed function in OpenSSL version 1.1
        2. in JNI side we have added flags to choose the right function based
           on installed version on Openssl
        3. Have updated JNA verion in pom.xml

Signed-off-by: ossdev  <ossdev@puresoftware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,15 +48,15 @@ $(TARGET)/jni-classes/org/apache/commons/crypto/OpenSslInfoNative.h: $(TARGET)/c
 
 $(COMMONS_CRYPTO_OUT)/OpenSslNative.o : $(SRC_NATIVE)/org/apache/commons/crypto/cipher/OpenSslNative.c $(TARGET)/jni-classes/org/apache/commons/crypto/cipher/OpenSslNative.h
 	@mkdir -p $(@D)
-	$(CC) $(CFLAGS) -c $< -o $@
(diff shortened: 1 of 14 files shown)


vanzin:
If I understand this correctly, you're still requiring different commons-crypto jars for different versions of OpenSSL, right?

If that's the case, sorry, but that has exactly the same problem I mentioned before.

The end goal here is that someone who pulls "org.apache.commons:commons-crypto" in their project can run that jar file against whatever version of OpenSSL they have locally. This is a real must have.

The secondary goal is that you should minimize the number of JNI libraries needed to be built. You can do the above by having different JNI libs for different versions of OpenSSL. But then, to release commons-crypto, you have to figure out how to build twice the number of native libraries as before.

Another choice is if JNA has become faster since last time people looked into it. I think it used to be super slow and couldn't really replace JNI for this use. But maybe that has changed...

ossdev07:
@vanzin if you are saying that two jars are not possible then there will a lot of duplicate code to provide support for both openssl  libraries and in my opinion it is not a good approach to keep duplicate code because in future openssl or any other libraries will be upgraded and for providing support we can not increase duplication of code.. if you have any idea/example please share as apart from above provided patches i don't have any optimal solution ..

vanzin:

PR #92, opened 2018-12-23
OpenSSL 1.1.0 updates with backward compatibility for OpenSSL 1.0.2 and 1.0.1

aremily:
Commons Crypto Dev Team,

Please consider this submission for incorporation into the master.  I believe it meets the criteria that Marcelo specified via the mailing list, in that it supports both 1.1.0, 1.0.2 and 1.0.1 with one jar file.  See the relevant thread at the url:

http://mail-archives.apache.org/mod_mbox/commons-dev/201811.mbox/browser

I made every effort to keep changes minimally invasive and to keep the the underlying OpenSSL library version transparent to the Java developer.  The JNA package dynamically determines whether the underlying OpenSSL library is 1.0.2 or 1.1.0; the Java developer needn't provide versioning information in any config file.  The JNI libraries, however, must be compiled against the specific version of OpenSSL being used, but once that is done there is no change to the Java API or requirement to provide version information as a config parameter.

This branch has been tested with Oracle Java 8 by running "mvn clean test" against the following platforms and versions of OpenSSL:

Ubuntu 16.04 and OpenSSL 1.0.2n
Ubuntu 18.04 and OpenSSL 1.1.0i
Mac High Sierra and OpenSSL 1.1.0i

I look forward to your feedback.

Alex


aremily:
Thanks for the feedback.  Let me think on it a bit more and I'll take it up
again once I've come up with an approach.  Perhaps ossdev07 would be
interested in collaborating, as it seems we're both trying to achive the
same objective.

aremily:
It's building the JNI library or libraries that I'll have to work out.  I
don't routinely code in C, so I need to do some research about how to build
multi-version support into one library, if that's even possible, or to
compile a separate JNI library for each supported version.

vanzin:
If you use runtime checks, you don't need to touch the build at all. The same JNI library should work everywhere. If you test on a 1.1 system, it should cover everything (since Travis here will test your changes on a 1.0.2 system).

If you use compile-time checks ("ifdefs"), then you need to build different libraries, which is why I prefer the approach above.

aremily:
Time got away from me over the holidays, but I finally got back to this
project.  Most of the conversion was indeed trivial, but between OpenSSL
version 1.0.X and 1.1.X the EVP_CIPHER_CTX structure was made opaque, such
that in 1.1.X the elements of the struct can only be accessed using the
appropriate accessor function.  My specific challenge is that I can't
compile context->encrypt in 1.1.X, and I can't find an
EVP_CIPHER_CTX_encrypting function in 1.0.X.  I've tried mirroring the
struct locally and casting it, but the structures didn't align properly and
"encrypt" didn't have a value of 1 or 0.  If you know of a straightforward
solution for sneaking a data type passed the compiler please share it.
Otherwise, I'm pondering a few ways forward, in no particular order.

1.  Pass the "encrypt" flag from the Java environment.  This would be very
invasive to the existing codebase and may even change the API.  Probably a
lot of refactoring.
2.  Build a separate library that contains the required functions and dlsym
them into the OpenSslNative.c object.  I think this would allow us to get
the context->encrypt reference through the 1.1.0 compiler, but we would
have to package another library with the distribution.  I'm also not sure
how much 1.0.X code we'd have to pull over into the legacy support library
in order to get all the required dependencies.
3.  Limit the functionality of the Commons Crypto library to avoid
context->encrypt.  The only place I can see that it's used is in the
AEADBadTagException checking.  Two of 113 tests error out.  Everything else
passes, but for that tiny snippet of code.  How important is the bad tag
checking?

Any advice?

vanzin:
If I understand `context->encrypt` correctly, it just tracks whether the context is being used for encryption or decryption, right?

If that's the case, you could stash that information in the `EVP_CTX_Wrapper` created by the commons-crypto library. Set it in `Java_org_apache_commons_crypto_cipher_OpenSslNative_init`, and change some internal functions to take the wrapper instead of the OpenSSL context directly.

Does that solve your problem?

aremily:
It does.  Thank you.  I ran the build successfully with 1.1.0 on Mac OSX
and 1.0.2 on Ubuntu 16.04, but when I pushed to GitHub Travis fails testing
because of a system crash that I can't reproduce.  Not sure how to
proceed.  Can you have a look?  I think it's what you're looking for in
terms of support to 1.0.2 and 1.1.0.

aremily:
Travis is running OpenSSL 1.0.1, correct?  I've been developing to 1.0.2
and 1.1.0.  I'm guessing that the build errors on Travis are version
related, so I'll stand up a 1.0.1 environment and see if I can recreate the
failure.

aremily:
Please review this most recent commit for compliance with the requirements for merging into the master branch.  I've ran mvn clean site against OpenSSL 1.0.1, 1.0.2 and 1.1.0.  All version handling is done at runtime.

.gitignore
```diff
@@ -1,37 +0,0 @@
-*~
```
vanzin: Don't delete this file.

.travis.yml
```diff
@@ -24,7 +24,6 @@ matrix:
         - "curl -L --cookie 'oraclelicense=accept-securebackup-cookie;'  http://download.oracle.com/otn-pub/java/jce/8/jce_policy-8.zip -o /tmp/policy.zip && sudo unzip -j -o /tmp/policy.zip *.jar -d `jdk_switcher home oraclejdk8`/jre/lib/security && rm /tmp/policy.zip"
       after_success:
         - mvn clean test jacoco:report coveralls:report
-
```
vanzin: Undo.

Makefile
```diff
@@ -18,7 +18,7 @@
 include Makefile.common
 
 COMMONS_CRYPTO_OUT:=$(TARGET)/$(commons-crypto)-$(os_arch)
-COMMONS_CRYPTO_OBJ:=$(addprefix $(COMMONS_CRYPTO_OUT)/,OpenSslCryptoRandomNative.o OpenSslNative.o OpenSslInfoNative.o)
+COMMONS_CRYPTO_OBJ:=$(addprefix $(COMMONS_CRYPTO_OUT)/,OpenSslInfoNative.o OpenSslCryptoRandomNative.o OpenSslNative.o)
```
vanzin: Nothing is changing here, right? If not, undo.

src/main/java/org/apache/commons/crypto/jna/OpenSslJnaCryptoRandom.java
```diff
@@ -103,10 +103,9 @@ public void nextBytes(byte[] bytes) {
             //to support multithreading https://wiki.openssl.org/index.php/Manual:Threads(3) needs to be done
             
             if(rdrandEnabled && OpenSslNativeJna.RAND_get_rand_method().equals(OpenSslNativeJna.RAND_SSLeay())) {
-                close();
```
vanzin: Why do you need to remove this?

src/main/native/org/apache/commons/crypto/OpenSslInfoNative.c
```diff
@@ -135,3 +96,35 @@ JNIEXPORT jstring JNICALL Java_org_apache_commons_crypto_OpenSslInfoNative_Nativ
...
     return (*env)->NewStringUTF(env, PROJECT_NAME);
 }
+
+JNIEXPORT jlong JNICALL Java_org_apache_commons_crypto_OpenSslInfoNative_OpenSSL
+  (JNIEnv *env, jclass clazz)
+{
+    if (!load_library(env)) {
+        return 0;
+    }
+    if(OPENSSL_VERSION_NUMBER > VERSION_1_1_X){
```
vanzin: nit: space after `if` (also in some other places)

Where does `OPENSSL_VERSION_NUMBER` come from? A quick google search seems to indicate it's a macro in an OpenSSL header. Which makes this a compile time check, basically.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
```diff
@@ -716,3 +677,52 @@ JNIEXPORT void JNICALL Java_org_apache_commons_crypto_cipher_OpenSslNative_clean
   EVP_CTX_Wrapper *wrapper = CTX_WRAPPER(ctx);
   free_context_wrapper(wrapper);
 }
+
+static int check_update_max_output_len(JNIEnv *env, EVP_CIPHER_CTX *context, jlong ctx, int input_len, int max_output_len)
```
vanzin: All these arguments feel a little redundant. e.g. you don't need `context` since you can just use `wrapper->context`.

With some small changes you could also just pass the wrapper directly from the callers, instead of passing the jni env + the pointer address.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
```diff
@@ -716,3 +677,52 @@ JNIEXPORT void JNICALL Java_org_apache_commons_crypto_cipher_OpenSslNative_clean
...
+static int check_doFinal_max_output_len(JNIEnv *env, EVP_CIPHER_CTX *context, int max_output_len)
+{
+  if (dlsym_EVP_CIPHER_CTX_test_flags(context, EVP_CIPH_NO_PADDING) == EVP_CIPH_NO_PADDING) {
+    return 1;
+  } else {
+    int b = dlsym_EVP_CIPHER_CTX_block_size(context);
+    if (max_output_len >= b) {
+      return 1;
+    }
+  return 0;
```
vanzin: indentation

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +278,57 @@ static unsigned long pthreads_thread_id(void)
  * If using an Intel chipset with RDRAND, the high-performance hardware
  * random number generator will be used.
  */
-static ENGINE * openssl_rand_init(void)
+static ENGINE * openssl_rand_init(JNIEnv *env)
 {
-  locks_setup();
+  if (OPENSSL_VERSION_NUMBER < VERSION_1_1_X) {
```
vanzin: Looks like another compile-time check, if `OPENSSL_VERSION_NUMBER` really comes from the openssl headers.

aremily:
Completely spaced off that that was a compile time check.  I'll swap it out
with a dynamic one.  What diff tool are you using?  I should clean up the
formatting stuff before I check it in.

vanzin:
Not using any tool, just reviewing the code here.

.gitignore
aremily: Added back in most recent commit.

.travis.yml
aremily: Undone.

Makefile
aremily: The build was order dependent.  It isn't anymore.

src/main/java/org/apache/commons/crypto/jna/OpenSslJnaCryptoRandom.java
aremily: I don't remember, but I added it back and tests pass fine.

src/main/native/org/apache/commons/crypto/OpenSslInfoNative.c
aremily: I replaced it with a dynamic lookup, following the existing pattern of dlsyming functions via a load macro.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
aremily: I replaced the env and address arguments with the wrapper argument in functions that required the wrapper.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
aremily: Fixed

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
aremily: Done.

aremily:
Please review this latest commit for incorporation into the master branch.  Thanks.

src/main/native/org/apache/commons/crypto/OpenSslInfoNative.c
```diff
@@ -94,30 +76,12 @@ static int load_library(JNIEnv *env)
     THROW(env, "java/lang/UnsatisfiedLinkError", msg);
     return 0;
   }
-  get_methods(env, openssl);
+#ifdef UNIX
```
vanzin: I'm sure we'll want this on Windows too at some point, but right now we need to figure out the Windows build in the first place, so this is ok.

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
...
 
-  dlsym_ENGINE_load_rdrand();
   ENGINE *eng = dlsym_ENGINE_by_id("rdrand");
 
-  int ret = -1;
-  do {
-    if (NULL == eng) {
-      break;
-    }
+    int ret = -1;
```
vanzin: nit: mind undoing these changes? Previous indentation was correct.

(Then we can see what's actually changing.)

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
...
+    return eng;
 }
 
-static void openssl_rand_clean(ENGINE *eng, int clean_locks)
+static void openssl_rand_clean(JNIEnv *env, ENGINE *eng, int clean_locks)
 {
   if (NULL != eng) {
-    dlsym_ENGINE_finish(eng);
-    dlsym_ENGINE_free(eng);
+	dlsym_ENGINE_finish(eng);
```
vanzin: nit: indentation

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
...
+	dlsym_ENGINE_free(eng);
   }
 
-  dlsym_ENGINE_cleanup();
-  if (clean_locks) {
-    locks_cleanup();
+  if(dlsym_OpenSSL_version_num() < VERSION_1_1_X) {
```
vanzin: nit: space after `if`

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
...
+	static void (*dlsym_ENGINE_cleanup) (void);
```
vanzin: nit: indentation in this whole block is wonky

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
...
+	dlsym_ENGINE_cleanup = do_dlsym(env, openssl, "ENGINE_cleanup");
```
vanzin: I know this probably will never fail, yet it would be nice to have a null check here.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
```diff
@@ -716,3 +676,50 @@ JNIEXPORT void JNICALL Java_org_apache_commons_crypto_cipher_OpenSslNative_clean
   EVP_CTX_Wrapper *wrapper = CTX_WRAPPER(ctx);
   free_context_wrapper(wrapper);
 }
+
+static int check_update_max_output_len(EVP_CTX_Wrapper *wrapper, int input_len, int max_output_len)
+{
+  if (dlsym_EVP_CIPHER_CTX_test_flags(wrapper->ctx, EVP_CIPH_NO_PADDING) == EVP_CIPH_NO_PADDING) {
+	if (max_output_len >= input_len) {
```
vanzin: Indentation should be 2 spaces, for consistency.

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -94,11 +94,30 @@ void *do_dlsym(JNIEnv *env, void *handle, const char *symbol) {
   return func_ptr;
 }
 
+static __attribute__ ((unused))
+void *do_version_dlsym(JNIEnv *env, void *handle) {
+  if (!env || !handle) {
+	  THROW(env, "java/lang/InternalError", NULL);
```
vanzin: indentation again

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -94,11 +94,30 @@ void *do_dlsym(JNIEnv *env, void *handle, const char *symbol) {
...
+	  return NULL;
+  }
+  void *func_ptr = dlsym(handle, "OpenSSL_version_num");
+  if (func_ptr == NULL) {
+	  func_ptr = dlsym(handle, "SSLeay");
```
vanzin: and here

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -94,11 +94,30 @@ void *do_dlsym(JNIEnv *env, void *handle, const char *symbol) {
...
 /* A helper macro to dlsym the requisite dynamic symbol and bail-out on error. */
 #define LOAD_DYNAMIC_SYMBOL(func_ptr, env, handle, symbol) \
   if ((func_ptr = do_dlsym(env, handle, symbol)) == NULL) { \
     return; \
   }
+
+/* A macro to dlsym the appropriate OpenSSL version number function. */
+#define LOAD_OPENSSL_VERSION_FUNCTION(func_ptr, env, handle) \
+if ((func_ptr = do_version_dlsym(env, handle)) == NULL) { \
+	THROW(env, "java/lang/Error", NULL); \
```
vanzin: and here; also the body of the macro should be indented a bit (see macro above)

src/main/native/org/apache/commons/crypto/random/OpenSslCryptoRandomNative.c
```diff
@@ -294,49 +280,57 @@ static unsigned long pthreads_thread_id(void)
  * If using an Intel chipset with RDRAND, the high-performance hardware
  * random number generator will be used.
  */
-static ENGINE * openssl_rand_init(void)
+static ENGINE * openssl_rand_init(JNIEnv *env)
 {
-  locks_setup();
+  if (dlsym_OpenSSL_version_num() < VERSION_1_1_X) {
+    locks_setup(env);
+	static void (*dlsym_ENGINE_load_rdrand) (void);
```
vanzin: again

aremily:
I think I got all the changes.  I finally changed my c file formatter to stop defaulting with tabs.  Should have done that a month ago.  Please have a look.

aremily:
Great.  When will I be able to pull the version with these changes out of the Maven repo?

vanzin:
Getting a release out will probably take a while, unless someone takes over fixing the Windows side of things, and figuring out how to properly create the release in the first place.

aremily:
The project I'm using commons crypto for makes heavy use of a keyed pseudorandom function, so eventually I'd like to add CMAC and/or HMAC functionality to commons crypto.  Thoughts?  Good idea?  Bad idea?

vanzin:
Either open a bug in jira, or send a message on the dev list, to start the conversation.

ossdev07:
@vanzin @aremily 
Please have a look on below findings 
I checked this patch on our AARCH64 platform (CentOS7) with openJDK8,  jvm is getting crashed every time
[ERROR] Error occurred in starting fork, check output in log
[ERROR] Process Exit Code: 134
[ERROR] Crashed tests:
[ERROR] org.apache.commons.crypto.jna.CtrCryptoJnaStreamTest
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?

When i changed _**<jna.version>4.2.2</jna.version>**_ with **<jna.version>4.4.0</jna.version>** in pom.xml it is running fine and all the test cases are getting passed

vanzin:
Then open a pull request.

aremily:
I'll have a look using OpenJDK.  To this point I've done everything on Oracle Java.  Will Travis support an OpenJDK build if I were to include it in the Travis.yml matrix?

vanzin:
https://docs.travis-ci.com/

(But really, discussing unrelated things in a closed PR is one of my pet peeves. Please use the mailing lists. Or jira.)

PR #94, opened 2019-04-11
Crypto 137 windows compile

aremily:
Please review for inclusion to the master branch.  Addresses CRYPTO-137 and in the process cleans up a lot of changes introduced in the 1.1 support branch merge.  Build has been tested and passes for OpenSSL versions 1.0 and 1.1 on Windows and Linux, and 1.1 on OSX.  Might also address CRYPTO-139.

Note that the Travis OSX build ships with LIbreSSL, and the LibreSSL build skips all the JNA tests.  Do we want to support LibreSSL?  If LibreSSL is the default brew install, it’s probably on a lot of Macs.  

The Travis Windows build does not support Java, so I omitted it.  The Travis Ubuntu build does not yet support Bionic with its default OpenSSL 1.1 install, but I would recommend adding a Bionic build to the Travis matrix as soon as it’s supported.  Currently all the Travis builds are on OpenSSL 1.0, so 1.1 builds must be tested locally.

Some of the line wrapping is inconsistent.  I prefer fewer breaks, so I made changes while I was troubleshooting that made the code easier for me to read.  I’d like to keep it that way, but understand that people can have strong feelings about it if you want it changed back.

Java 8 is the last version that supports “javah”.  Subsequent versions will require javac -h, which does not support the same command line options as javah.  Something to consider if we want to support more recent releases of Java.


aremily:
As you can see from the build history, the testGcmTamperedData() test, which expects an AEADBadTagException, fails intermittently on the Travis Linux build.  I can't seem to recreate it in any other environment.  It fails by throwing an internal error as opposed to an AEADBadTagException, so I'm not sure how big of a deal it really is.  It's not like it doesn't fail when it's supposed to, it just doesn't fail in the way it's supposed to.  Advice welcome.

src/main/native/org/apache/commons/crypto/OpenSslInfoNative.c
```diff
@@ -41,27 +41,48 @@
...
+static void get_methods(JNIEnv *env, void *openssl)
+#endif
+#ifdef WINDOWS
+static void get_methods(JNIEnv *env, HMODULE openssl)
+#endif
+{
+  LOAD_OPENSSL_VERSION_FUNCTION(dlsym_OpenSSL_version_num, env, openssl);
+#ifdef UNIX
+  if (dlsym_OpenSSL_version_num() > VERSION_1_1_X) {
+      LOAD_DYNAMIC_SYMBOL(dlsym_OpenSSL_version, env, openssl, "OpenSSL_version");
```
vanzin: 2 vs. 4 space indent, here and in other places in this file.

src/main/native/org/apache/commons/crypto/cipher/OpenSslNative.c
```diff
@@ -183,12 +188,18 @@ JNIEXPORT void JNICALL Java_org_apache_commons_crypto_cipher_OpenSslNative_initI
...
-  LOAD_DYNAMIC_SYMBOL(__dlsym_EVP_CIPHER_CTX_reset,  \
-                      dlsym_EVP_CIPHER_CTX_reset, env,
-                      openssl, "EVP_CIPHER_CTX_reset");
-  LOAD_DYNAMIC_SYMBOL(__dlsym_EVP_CIPHER_CTX_set_padding,  \
-                      dlsym_EVP_CIPHER_CTX_set_padding, env,  \
-                      openssl, "EVP_CIPHER_CTX_set_padding");
+  LOAD_DYNAMIC_SYMBOL(__dlsym_EVP_CIPHER_CTX_set_padding, dlsym_EVP_CIPHER_CTX_set_padding,  \
+                      env, openssl, "EVP_CIPHER_CTX_set_padding");
+  LOAD_DYNAMIC_SYMBOL(__dlsym_EVP_CIPHER_CTX_ctrl, dlsym_EVP_CIPHER_CTX_ctrl,  \
+		              env, openssl, "EVP_CIPHER_CTX_ctrl");
```
vanzin: alignment is inconsistent with previous calls (also happens below in several calls)

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -134,7 +134,7 @@ void *do_version_dlsym(JNIEnv *env, void *handle) {
 
 /* Microsoft C Compiler does not support the C99 inline keyword */
 #ifndef __cplusplus
-#define inline __inline;
+//#define inline __inline;
```
vanzin: What is the problem here? If this is broken then perhaps the whole thing should be deleted, instead of this line commented out.

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -187,6 +187,24 @@ static FARPROC WINAPI do_dlsym(JNIEnv *env, HMODULE handle, LPCSTR symbol) {
   }
   return func_ptr;
 }
+
+static FARPROC WINAPI do_version_dlsym(JNIEnv *env, HMODULE handle) {
+  FARPROC func_ptr = NULL;
+  if (!env || !handle) {
+    THROW(env, "java/lang/InternalError", NULL);
+      return NULL;
```
vanzin: indentation

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
```diff
@@ -187,6 +187,24 @@ static FARPROC WINAPI do_dlsym(JNIEnv *env, HMODULE handle, LPCSTR symbol) {
...
+  }
+  func_ptr = GetProcAddress(handle, "OpenSSL_version_num");
+  if (func_ptr == NULL) {
+    func_ptr = GetProcAddress(handle, "SSLeay");
+  }
+  return func_ptr;
+}
+/* A macro to dlsym the appropriate OpenSSL version number function. */
```
vanzin: Add empty line before.

src/main/native/org/apache/commons/crypto/org_apache_commons_crypto.h
aremily: I had to comment out that line to get it to compile with MinGW-64.  Redefining the keyword causes a cascade of errors.  I left it commented as opposed to removing it completely because presumably some developer had to redefine inline to get it to compile under some other Microsoft compiler, and perhaps it would be helpful to keep it there for reference.

aremily:
Thanks for taking the time to review my PR.  I think I addressed all the points you made, although I opted to leave the inline definition comment alone for reasons listed above.  Travis tested the OpenSSL 1.0 API on Linux.  Travis' Bionic distribution doesn't support Java 8, so I've attached my local build log for OpenSSL 1.1 on Linux.  I've also attached my local build log for OpenSSL 1.0 and 1.1 on Windows. The Travis Mac build uses LibreSSL, so I've attached my local build log for OpenSSL 1.1.  I haven't tested OpenSSL 1.0 on a Mac.

[Mac OSX Java 8 OpenSSL 1_1 Build Console Output.txt](https://github.com/apache/commons-crypto/files/3512616/Mac.OSX.Java.8.OpenSSL.1_1.Build.Console.Output.txt)
[Ubuntu Bionic Java 8 OpenSSL 1_1 Build Console Output.txt](https://github.com/apache/commons-crypto/files/3512617/Ubuntu.Bionic.Java.8.OpenSSL.1_1.Build.Console.Output.txt)
[Windows 7 Java 8 OpenSSL 1_0 Build Console Output.txt](https://github.com/apache/commons-crypto/files/3512618/Windows.7.Java.8.OpenSSL.1_0.Build.Console.Output.txt)
[Windows 7 Java 8 OpenSSL 1_1 Build Console Output.txt](https://github.com/apache/commons-crypto/files/3512619/Windows.7.Java.8.OpenSSL.1_1.Build.Console.Output.txt)

aremily:
I checked in my edits.  I think I caught everything.  I added the console
output from the test runs in the different environments to assist with your
review.

Thanks again,

Alex

aremily:
I saw that, too.  I don't believe that I introduced that behavior.  I can't
be certain, but I think this is an open issue:

https://issues.apache.org/jira/browse/CRYPTO-133

I took pains not to materially modify the locking code during my
refactoring.  I'll have another look at it if you disagree, but at this
point I'd just recommend synchronizing the calling method.

Thoughts?

PR #96, opened 2020-04-18
CI Improvements

adamretter:
1. Tidies up the Travis CI configuration
2. Adds Arm64 builds to Travis CI
3. Adds ppc64le builds to Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,23 @@ language: java
 
 matrix:
   include:
-    - name: "Ubuntu 14.04 / Java 8 / OpenSSL 1.0.x"
+    - name: "x64 / Ubuntu 14.04 / Java 8 / OpenSSL 1.0.x"
+      arch: amd64
       os: linux
       dist: trusty
       before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -18,28 +18,84 @@ language: java
 
 matrix:
   include:
-    - name: "Ubuntu 14.04 / Java 8 / OpenSSL 1.0.x"
+
+    - name: "x64 / Ubuntu 14.04 / Java 8 / OpenSSL 1.0.x"
```
geoffreyblake, line 22: Should this be the same as ppc64le and aarch64 build and be Ubuntu 16.04?

adamretter: I left it as 14.04 as that is what it was before. An arguable advantage of using an older Ubuntu is that it will have an older GCC so you can ensure wider compatibility.

Unfortunately 14.04 isn't available for ppc64le and aarch64 on Travis, which is why I used 16.04 instead.

geoffreyblake: I don't fully agree, but this is still a minor nit.  Probably fine to keep as is.

adamretter: Okay no worries. Could you explain? I am happy to change it if it makes sense...

vanzin:
Travis looks pretty green, so looks good to me. Merging.

PR #98, opened 2020-04-21
Update pom.xml to explicitly prevent cleaning during release:prepare

geoffreyblake:
Update pom.xml to configure the maven-release-plugin to not perform and clean when doing a release:prepare to allow build artifacts from other platform builds to be incorporated into the final Jar for publishing to the maven artifact repositories.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -606,6 +606,14 @@ The following provides more details on the included cryptographic software:
           <ignoredDifferencesFile>${basedir}/clirr-excludes.xml</ignoredDifferencesFile>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-release-plugin</artifactId>
+        <version>2.5.3</version>
+        <configuration>
+          <preparationGoals>verify</preparationGoals>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
   <reporting>


pom.xml
```diff
@@ -606,6 +606,14 @@ The following provides more details on the included cryptographic software:
           <ignoredDifferencesFile>${basedir}/clirr-excludes.xml</ignoredDifferencesFile>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-release-plugin</artifactId>
+	      <version>2.5.3</version>
+	      <configuration>
+          <preparationGoals>verify</preparationGoals>
+	      </configuration>
```
vanzin, line 615: Indentation is funky.

geoffreyblake: My text editor for some reason is mixing tabs and spaces.  Should be fixed now.

geoffreyblake:
Any comments on this?  If building by hand, this seems to be a way for making the combined jar.  I've come across another example: https://github.com/theopenlab/leveldbjni shows a way to build all the artifacts and then combine them, but it requires major surgery to the pom files.

geoffreyblake:
Any opinion on this PR? Or do we need something more sophisticated to publish and release the jar for multi-arch?

vanzin:
Gary didn't comment, and I'm ok with it, so merging.

PR #99, opened 2020-04-26
JaCoCo Increase for Streams

aremily:
Increases the stream package code coverage to 90% and the overall project coverage to 80%.


src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,236 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+        } catch (IOException ex) {
+            Assert.assertNotNull(ex);
+        }
+        
+        try { // Test closed stream.
+            in = getCryptoInputStream(new ByteArrayInputStream(encData), 
+                    getCipher(cipherClass), defaultBufferSize, iv, withChannel);
+            in.close();
+            in.close(); // Don't throw exception.
+            in.read(); // Throw exception.
```
geoffreyblake, line 391: Why 2 calls to close()?

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,236 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+        } catch (IOException ex) {
+            Assert.assertTrue(ex.getMessage().equals("Stream closed"));
+        } 
+
+        try { // Test closed stream.
+            out = getCryptoOutputStream(transformation, props, baos, key, new IvParameterSpec(iv), 
+                    withChannel);
+            out.close();
+            out.close(); // Don't throw exception.
```
geoffreyblake, line 400: Same question here.

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
```diff
@@ -52,4 +75,115 @@ protected CtrCryptoOutputStream getCryptoOutputStream(
...
+                return; // Skip this test if no JNI
+            }
+        }
+        
+        StreamInput streamInput = new StreamInput(new ByteArrayInputStream(encData), 0);
+        try {
+            streamInput.seek(0);
+            Assert.fail("Expected UnsupportedOperationException.");
+        } catch (UnsupportedOperationException ex) {
+        	Assert.assertEquals(ex.getMessage(), "Seek is not supported by this implementation");
```
geoffreyblake, line 104: Isn't asserting the exception is generated enough?  The same goes for the below tests as well, why compare the error message?

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: I needed to close a closed stream to hit the return branch and cover the line.  
  public void close() throws IOException {
        if (closed) {
            return;
        }

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: Same answer.

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
aremily: In this case I believe it would be enough.  IIRC, I was following an existing pattern for other tests which covered cases where different exceptions were caught, wrapped into a common exception and thrown as that more general exception, which was then only differentiated by the message.  I don't think it's hurting anything but I have no issue with you removing it.

aremily: Actually, now that I think about it I'd prefer it remained in.  I think it helps with readability.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
geoffreyblake: I see. Then should we split this test into 2 try blocks?  If someone removes that path you're testing with the 2nd close, the test will still pass because you'd catch an IOException from the 2nd close failing instead of from the read() call.

To be honest, I'd expect a double close() to throw and exception, but obviously not in this case.

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
geoffreyblake: I agree it helps with readability.  My reservation is this is a brittle test if the generation method changes in the future, all these tests have to be changed.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: Interesting point.  I did some digging.  In that scenario, if an IOException were thrown at the second call to close, the test would fail at line 393 unless the exception thrown happened to have the same message.  

Nevertheless, I ran the tests with the branch in question removed and the test passes because the second close doesn't throw an exception at all (I stepped through it).  The method contract forces us to deal with an IOException, but it turns out that one isn't thrown in the existing implementation.

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
aremily: I agree, but I think that's the point of the tests.  I would want a future developer to be conscious of the impact that changes had on the test and make necessary modifications, but I don't want to get into a philosophical discussion about software testing...

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
geoffreyblake: Can you still split the test into 2?  One to test double close(), and one to test operating on a closed stream?  That way it is explicit there are two cases being tested here.

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
geoffreyblake: Lets keep it as is then.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
geoffreyblake: LGTM

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
geoffreyblake: LGTM

geoffreyblake:
LGTM.  Need @vanzin to merge though.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
                 getCipher(cipherClass), smallBufferSize, iv, withChannel);
         buf.clear();
         byteBufferReadCheck(in, buf, 11);
+        in.close();      
+        
+        // Direct buffer, default buffer size, initial buffer position is 0, final read
```
vanzin, line 229: This says default but code says `smallBufferSize`. Not sure if that matters... also looking at the code you're adding, it seems a bit redundant with existing code, but didn't really double-check things.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+        try { // Test InvalidAlgorithmParameters
+        	in = getCryptoInputStream(transformation, props, new ByteArrayInputStream(encData), 
+                    new SecretKeySpec(key, "AES"), new GCMParameterSpec(0, new byte[0]), 
+                    withChannel);
+            Assert.fail("Expected IOException.");
+        } catch (IOException ex) {
+            Assert.assertEquals(ex.getMessage(),"Illegal parameters");
+        } 
+        
+        try { // Test InvalidAlgorithmParameters
```
vanzin, line 361:
```suggestion
        // Test InvalidAlgorithmParameters
        try {
```

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+            final boolean withChannel) throws IOException {
+        if (AbstractCipherTest.OPENSSL_CIPHER_CLASSNAME.equals(cipherClass)) {
+            if (!Crypto.isNativeCodeLoaded()) {
+                return; // Skip this test if no JNI
+            }
+        }
+        
+        InputStream in = null;
+        OutputStream out = null;
+        try { // Test InvalidAlgorithmParameters
```
vanzin, line 352:
```suggestion
        // Test InvalidAlgorithmParameters
        try {
```

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+            out = getCryptoOutputStream(transformation, props, baos, 
+                    new SecretKeySpec(key, "AES"), new GCMParameterSpec(0, 
+                    new byte[0]), withChannel);
+            Assert.fail("Expected IOException.");
+        } catch (IOException ex) {
+        	Assert.assertEquals(ex.getMessage(),"Illegal parameters");
+        } 
+        
+        try { // Test Invalid Key
```
vanzin, line 370:
```suggestion
        // Test Invalid Key
        try {
```

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+            in = getCryptoInputStream(transformation,props, new ByteArrayInputStream(encData), 
+                    new SecretKeySpec(new byte[10], "AES"), new IvParameterSpec(iv), withChannel);
+            Assert.fail("Expected IOException for Invalid Key");
+        } catch (IOException ex) {
+            Assert.assertNotNull(ex);
+        }
+        
+        try { // Test Invalid Key
```
vanzin, line 378:
```suggestion
        // Test Invalid Key
        try {
```

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+            out = getCryptoOutputStream(transformation, props, baos, new byte[10], 
+                    new IvParameterSpec(iv), withChannel);
+            Assert.fail("Expected IOException for Invalid Key");
+        } catch (IOException ex) {
+            Assert.assertNotNull(ex);
+        }
+        
+        try { // Test reading a closed stream.
```
vanzin, line 386: I think you get the idea... (the suggestion editor is not very good so adding those is kinda painful.)

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+            Assert.assertTrue(ex.getMessage().equals("AES/CTR/NoPadding is required"));
+        } finally {
+            in.close();
+        }
 
-        out.flush();
+        try { // Test unsupported operation handling.
+            in = getCryptoInputStream(new ByteArrayInputStream(encData), 
+                    getCipher(cipherClass), defaultBufferSize, iv, false);
+            in.mark(0); // Should not throw an exception.
```
vanzin, line 427: Normally things that should not throw an exception should be outside the `try`. (You're also leaving the stream open but that's minor in a test.)

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,246 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
+        Assert.assertEquals(in.getBufferSize(), defaultBufferSize);
+        Assert.assertEquals(in.getCipher().getClass(), Class.forName(cipherClass));
+        Assert.assertEquals(in.getKey().getAlgorithm(), "AES");
+        Assert.assertEquals(in.getParams().getClass(), IvParameterSpec.class);
+        Assert.assertNotNull(in.getInput());
+
+        CryptoOutputStream out = getCryptoOutputStream(baos, getCipher(cipherClass), 
+                defaultBufferSize, iv, withChannel);
 
+        Assert.assertEquals(out.getOutBuffer().capacity(), defaultBufferSize + 16);
```
vanzin, line 462: Is there a constant for the `16` anywhere?

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -282,6 +570,24 @@ protected CryptoInputStream getCryptoInputStream(final ByteArrayInputStream bais
         return new CryptoInputStream(bais, cipher, bufferSize,
                 new SecretKeySpec(key, "AES"), new IvParameterSpec(iv));
     }
+    
+    protected CryptoInputStream getCryptoInputStream(final String transformation, final Properties props, 
+    	    final ByteArrayInputStream bais, final byte[] key, final AlgorithmParameterSpec params,
+    	    boolean withChannel) throws IOException {
+        if(withChannel) {
```
vanzin, line 577:
```suggestion
        if (withChannel) {
```

But can't you just call the other `getCryptoInputStream`?

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -294,6 +600,26 @@ protected CryptoOutputStream getCryptoOutputStream(
         return new CryptoOutputStream(baos, cipher, bufferSize,
                 new SecretKeySpec(key, "AES"), new IvParameterSpec(iv));
     }
+    
+    protected CryptoOutputStream getCryptoOutputStream(final String transformation,
```
vanzin, line 604: can just call the other `getCryptoOutputStream`?

src/test/java/org/apache/commons/crypto/stream/CtrCryptoStreamTest.java
```diff
@@ -52,4 +75,115 @@ protected CtrCryptoOutputStream getCryptoOutputStream(
...
+    
+    @Override
+    protected CtrCryptoOutputStream getCryptoOutputStream(final String transformation,
+            final Properties props, final ByteArrayOutputStream baos, final byte[] key, 
+            final AlgorithmParameterSpec params, final boolean withChannel) throws IOException {
+        if (withChannel) {
+            return new CtrCryptoOutputStream(props, Channels.newChannel(baos), key, 
+                    ((IvParameterSpec)params).getIV());
+        }
+        return new CtrCryptoOutputStream(props, baos, key,((IvParameterSpec)params).getIV());
```
vanzin, line 87:
```suggestion
        return new CtrCryptoOutputStream(props, baos, key, ((IvParameterSpec)params).getIV());
```

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: Nice catch.  The comment is a copy-paste oversight and I'll fix it.  I agree that these tests are largely redundant.  I needed to run different constructors and branch into some final logic in the read method to increase the code coverage.  I added a test that forced the read method to branch.  Then I just ran the tests again using a getCryptoInputSteam method that called different constructors.  It is a bit overkill, but I don't think it does any harm to run each test again against a separate constructor, as opposed to simply picking one test at random to cover the additional constructors.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: I'll close the stream.  I snuck the mark method in here only to add to coverage, and it seemed like the most logical place to put it.  I'm really testing that mark and reset are not supported.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: I'll replace it with getBlockSize.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
aremily: Unused constructors were a significant gap in code coverage, so it was an easy win to overload this method and use it to rerun existing tests against them.  Same with the getInputStream, as alluded to in an earlier comment.

aremily:
The test timeouts were in the original codebase, so I just added them to the new tests to be consistent.

aremily:
@ Vanzin--I think I hit all your points.  Can you have another look at this
PR?

Alex

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
vanzin: Wait, you're actually expecting this exception, at least now, so the comment is wrong. I'll remove the comment by using the suggestion thing and then merge.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -197,46 +224,261 @@ protected void doByteBufferRead(final String cipherClass, final boolean withChan
...
-        try (InputStream in = getCryptoInputStream(
-                new ByteArrayInputStream(encData), getCipher(cipherClass),
-                defaultBufferSize, iv, withChannel)) {
-            buf = ByteBuffer.allocate(dataLen + 100);
-            byteBufferReadCheck(in, buf, 0);
+        // Test unsupported operation handling.
+        try { 
+            in = getCryptoInputStream(new ByteArrayInputStream(encData), 
+                    getCipher(cipherClass), defaultBufferSize, iv, false);
+            in.mark(0); // Should not throw an exception.
```
vanzin, line 438:
```suggestion
            in.mark(0);
```

vanzin:
Test failure looks like the GCM test that I've seen mentioned as being flaky. Previous run had passed in any case. Merging.

PR #103, opened 2020-08-02
Standardize the release build process with Docker.

aremily:
Following up on Marcelo's recommendation to Dockerize the build process, submitted for review is a Dockerfile that automates the following builds:

Linux x86_64
Linux-arm
Linux-armfh
Linux-aarch64
Windows x86_64

This Dockerfile copies the project directory into the Docker image, so if you run it from a Mac after a successful local build, the resulting image will contain a jar that includes the above builds and the Mac build.  I've tested the Linux x86_64, Windows x86_64 and Mac x86_64 builds.  I have not tested the ARM builds, as I lack the platform to do so.

Note that this doesn't automate the Mac build, but Google suggests that it is possible.  I'll make the effort it if there's interest.

https://github.com/tpoechtrager/osxcross
https://github.com/multiarch/crossbuild

I included a Maven profile that runs the Dockerfile from Maven after a local build, but the plugin doesn't include much logging.  I'd recommend just running the Dockerfile from the command line, which will give you plenty of log output.

I don't pretend to be a Docker wizard, so feedback is appreciated.  There's probably a more efficient way to do this if anyone wants to give it a go.  The build process takes between ten and fifteen minutes on my machine because I build OpenSSL from source for each architecture to generate the platform-specific opensslconf.h file that's required for each build.  I'm not completely sure that it's necessary for purposes of building the JNI libraries, so if that step can be eliminated it would speed up the build considerably.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,16 +20,19 @@
 
 FROM ubuntu:18.04
 ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+# Install dependencies and tooling.
 RUN dpkg --add-architecture i386 && apt-get update && apt-get --assume-yes install build-essential \
       && apt-get --assume-yes install openjdk-8-jdk && apt-get --assume-yes install maven \
       && apt-get --assume-yes install libssl-dev:i386 && apt-get --assume-yes install libssl-dev \
       && apt-get --assume-yes install gcc-arm-linux-gnueabi && apt-get --assume-yes install g++-arm-linux-gnueabi \
       && apt-get --assume-yes install gcc-arm-linux-gnueabihf && apt-get --assume-yes install g++-arm-linux-gnueabihf \
       && apt-get --assume-yes install gcc-aarch64-linux-gnu && apt-get --assume-yes install g++-aarch64-linux-gnu \
-      && apt-get --assume-yes install mingw-w64 && apt-get --assume-yes install wget && apt-get --assume-yes install curl \
-      && mkdir commons-crypto && cp /usr/include/x86_64-linux-gnu/openssl/opensslconf.h /usr/include/openssl
+      && apt-get --assume-yes install mingw-w64 \
+# Copy the opensslconf.h file to the base openssl include directory
+      && cp /usr/include/x86_64-linux-gnu/openssl/opensslconf.h /usr/include/openssl \
+# Create the build directory.
+      && mkdir commons-crypto 
 COPY . /commons-crypto
-RUN cd commons-crypto && mvn package
-# Build openssl from source to generate the platform-specific opensslconf.h for the cross-compilers
-RUN cd commons-crypto && mvn -DskipTests package -P linux-arm && mvn -DskipTests package -P linux-armhf 
-RUN cd commons-crypto && mvn -DskipTests package -P linux-aarch64 && mvn -DskipTests package -P win64
+# Run the base Linux x86_64 build with tests and then run the cross-compile builds without.
+RUN cd commons-crypto && mvn package && mvn -DskipTests package -P linux-arm && mvn -DskipTests package -P linux-armhf \
+      && mvn -DskipTests package -P linux-aarch64 && mvn -DskipTests package -P win64
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,26 +13,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This file builds the Linux-x86_64, Linux-arm, Linux-armfh. Linux aarch64 and Win64 jnilibs.  It copies the contents of
-# the build host's project directory (commons-crypto) into the docker image and cross compiles the remaining builds. If 
-# you run this script from a Mac after a successful build, the build in the resuing Docker image will include the Mac 
-# build.
+# This file runs builds for the Linux-x86_64, Linux aarch64, Linux-arm, Linux-armfh and Win64 
+# architectures.  It copies the contents of the build host's project directory (commons-crypto) 
+# into the docker image, builds and tests the x86_64 build natively, and then cross compiles the 
+# remaining builds.  If you run this script from a Mac after a successful build, the build in the 
+# resuing Docker image will also include the Mac build by virtue of the initial project directory 
+# copy.
 
-FROM ubuntu:18.04
+FROM ubuntu:14.04
 ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
-# Install dependencies and tooling.
-RUN dpkg --add-architecture i386 && apt-get update && apt-get --assume-yes install build-essential \
-      && apt-get --assume-yes install openjdk-8-jdk && apt-get --assume-yes install maven \
-      && apt-get --assume-yes install libssl-dev:i386 && apt-get --assume-yes install libssl-dev \
-      && apt-get --assume-yes install gcc-arm-linux-gnueabi && apt-get --assume-yes install g++-arm-linux-gnueabi \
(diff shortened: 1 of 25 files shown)


aremily:
Looks like some of our Travis builds need work.  Half the builds are failing for environmental reasons unrelated to the application.

garydgregory:
I thought there were -dev packages that contain openssl C headers that one can install with apt-get?

aremily:
There are.  The script installs them and they are used for the Linux x86_64
build.  While developing this script I discovered that one included header
file, opensslconf.h, is platform specific and generated by Make during the
OpenSSL build process.  Unsure whether or not using the opensslconf.h file
from the x86_64 build would matter to the JNI libraries for the other
platforms, I decided to cross-compile OpenSSL to the Commons Crypto build's
target platform and use the opensslconf.h specific to that platform for
each build.  If that step is unnecessary it would probably shave at least
ten minutes off the build.

garydgregory:
On Sun, Aug 2, 2020 at 1:55 PM aremily <notifications@github.com> wrote:

> There are. The script installs them and they are used for the Linux x86_64
> build. While developing this script I discovered that one included header
> file, opensslconf.h, is platform specific and generated by Make during the
> OpenSSL build process. Unsure whether or not using the opensslconf.h file
> from the x86_64 build would matter to the JNI libraries for the other
> platforms, I decided to cross-compile OpenSSL to the Commons Crypto build's
> target platform and use the opensslconf.h specific to that platform for
> each build. If that step is unnecessary it would probably shave at least
> ten minutes off the build.
>

When Geoffrey Blake helped me set up an Ubuntu 14.04 docker image, we never
cross-compiled OpenSSL and we got the following working:

mvn -V -DskipTests package -Plinux-linux64
mvn -V -DskipTests package -Plinux-linux32
mvn -V -DskipTests package -Plinux-win64
mvn -V -DskipTests package -Plinux-win32
mvn -V -DskipTests package -Plinux-aarch64


For the Maven profile linux-aarch64, installing the proper
g++-4.8-aarch64-linux-gnu removed packages cc-multilib and g++-multilib
such that you had to be done building other binaries before aarch64.

See my email from earlier this week about what to do for Mac binaries.

Gary

aremily:
Thanks for the response, Gary.  I know you've been busy lately with all the
builds I've seen going out and I appreciate your time.  I'll modify the
script to use the opensslconf.h from the base system for all the cross
builds.

Regarding the Mac build, the only related email I've seen is one from you
to the community inquiring into the process of obtaining binaries from
outside sources.  If the Mac build is still an open issue I'll look into
including it in the Dockerfile.

Alex

garydgregory:
@aremily 
For the next release 1.1.0, I'd like to maintain the same binaries we shipped in 1.0.0 but I am not sure that is possible since that included mac32 and mac64 and it is not clear to me if that's possible on Docker. Including aarch64 would be new for 1.1.0.

aremily:
Gary,

I'll take a look at Dockerizing the Mac64 build.  Do you have any
heartburn with omitting the Win32 and Mac32 builds?  I'm not sure
they're relevant anymore.

Alex

garydgregory:
Well, I would be ok with that.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file builds the Linux-x86_64, Linux-arm, Linux-armfh. Linux aarch64 and Win64 jnilibs.  It copies the contents of
+# the build host's project directory (commons-crypto) into the docker image and cross compiles the remaining builds. If 
+# you run this script from a Mac after a successful build, the build in the resuing Docker image will include the Mac 
+# build.
+
+FROM ubuntu:18.04
```
geoffreyblake, line 21: For building binaries to be used in a release, this should be ubuntu:14.04 or even centos:7 to retain as much compatibility as possible.

geoffreyblake: If we want to simplify the docker files to not cross-compile, we could also add this to the FROM clause: `--platform=linux/aarch64`

geoffreyblake:
I brought up an x86 and aarch64 container and looked at the opensslconf.h's for both.  There appears to be no difference that should affect commons-crypto being able to bind to either OpenSSL 1.1.x or OpenSSL 1.0.x.   I do not think @aremily that you need to build OpenSSL from source.

On the Mac front, considering the license for MacOS prohibits running it on non-Apple hardware, a docker image for it being committed to this repository may be on shaky ground.

aremily:
That being the case the script is basically done.  I'll do a little cleanup
and check in another version, but unless there are other architectures to
include it's come about as far as it can go.

garydgregory:
@aremily 
May you please rebase on master?

Dockerfile
aremily: 14.04 is end of life, so I went with the most current LTS version of Ubuntu.  I chose Ubuntu because the comments in the makefile specified the proper packages for cross-compiling.  I opted for the cross-compile approach as opposed to pulling down a new base image for each platform so that each iteration would add an architecture to the existing build.  If I built each architecture on a separate base image I'd have to do a lot of build-dir copying between images.

aremily:
I think I just did.

Dockerfile
geoffreyblake: Sure, though for purposes of a release, it'd be best to take a distro that is 'old'.  The main concern is libc compatibility as compiling against newer versions **is not** backward compatible to older versions, whereas compiling against an old version of libc **is** forward compatible.  

Ubuntu 14.04 is still supported by Canonical according to their Docker hub info page, and is old enough for Arm builds, but for x86, it'd be better to go further back to something like CentOS7 or even 6.

garydgregory: FWIW, what I have working now for my local builds is based on 14.04 which is what I'll use to cut a release candidate hopefully soon(ish).

aremily:
Fair enough.  I'll make the update.

Makefile.common
```diff
@@ -163,17 +163,17 @@ Linux-arm_COMMONS_CRYPTO_FLAGS:=
...
 Linux-armhf_LIBNAME   := libcommons-crypto.so
 Linux-armhf_COMMONS_CRYPTO_FLAGS:=
 
 Linux-aarch64_CC        := $(CROSS_PREFIX)gcc
 Linux-aarch64_CXX       := $(CROSS_PREFIX)g++
 Linux-aarch64_STRIP     := $(CROSS_PREFIX)strip
-Linux-aarch64_CXXFLAGS  := -Ilib/inc_linux -I"$(JAVA_HOME)/include" -Ilib/inc_mac -O2 -fPIC -fvisibility=hidden -Wall -Werror
-Linux-aarch64_CFLAGS    := -Ilib/inc_linux -I"$(JAVA_HOME)/include" -Ilib/inc_mac -O2 -fPIC -fvisibility=hidden -Wall -Werror
+Linux-aarch64_CXXFLAGS  := -I"$(JAVA_HOME)/include" -I/usr/include/i386-linux-gnu -O2 -fPIC -fvisibility=hidden -Wall -Werror
+Linux-aarch64_CFLAGS    := -I"$(JAVA_HOME)/include" -I/usr/include/i386-linux-gnu -O2 -fPIC -fvisibility=hidden -Wall -Werror
```
geoffreyblake, line 176: Some documentation in here may be required to explain why there's an i368 architecture directory specified in the aarch64 compile.

aremily:
Let me add that to the list of things to look at.  It may be something I
forgot to go back and change.

PR #127, opened 2021-01-05
[CRYPTO-156] - Common Class Padding and Transform

arturobernalg:
In order to avoid duplicate code and try to unify the transformation of the token i think it's necessary create the next class:

OpenSslTransform --> Utility code for dealing with different algorithm types
OpenSslPadding --> Containg the enumeration of Cipher Algorithm Padding
OpenSslAlgorithmMode --> Enumeration of Algorithm Mode.
https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/cipher/OpenSsl.java#L208
https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/jna/OpenSslJnaCipher.java#L422
https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/cipher/OpenSsl.java#L47
https://github.com/apache/commons-crypto/blob/master/src/main/java/org/apache/commons/crypto/jna/OpenSslJnaCipher.java#L399

### DIFF
--- a/src/main/java/org/apache/commons/crypto/OpenSslTransform.java
+++ b/src/main/java/org/apache/commons/crypto/OpenSslTransform.java
@@ -1,0 +1,77 @@
+ /*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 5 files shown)


aremily:
Any committer want to have a look at this?  Seems to me like a code improvement.

garydgregory:
> Any committer want to have a look at this? Seems to me like a code improvement.

I'll take a look this weekend. Nothing urgent IMO since it should all be private/internal.

PR #128, opened 2021-01-07
Fix PMD warning and don't init to defaults

arturobernalg:
* Remove Redundant Field Initializer
* Use Files.newInputStream(Paths.get(fileName))
* Remove Instantiating Objects In Loops

### DIFF
--- a/src/main/java/org/apache/commons/crypto/NativeCodeLoader.java
+++ b/src/main/java/org/apache/commons/crypto/NativeCodeLoader.java
@@ -19,7 +19,6 @@ package org.apache.commons.crypto;
 
 import java.io.BufferedInputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
@@ -144,8 +143,9 @@ final class NativeCodeLoader {
                 return null;
             }
             // Extract a native library file into the target directory
+            final Path path;
             try {
-                final Path path = extractedLibFile.toPath();
+                path = extractedLibFile.toPath();
                 final long byteCount = Files.copy(inputStream, path, StandardCopyOption.REPLACE_EXISTING);
                 if (isDebug()) {
                     debug("Extracted '%s' to '%s': %,d bytes [%s]", nativeLibraryFilePath, extractedLibFile, byteCount,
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java
```diff
@@ -587,10 +587,10 @@ static void freeDirectBuffer(final ByteBuffer buffer) {
             .clean(); */
             final String SUN_CLASS = "sun.nio.ch.DirectBuffer";
             final Class<?>[] interfaces = buffer.getClass().getInterfaces();
+            final Object[] NO_PARAM = new Object[0];
```
garydgregory, line 590: You can use the same name we have in Commons Lang: EMPTY_OBJECT_ARRAY.

src/main/java/org/apache/commons/crypto/NativeCodeLoader.java
```diff
@@ -171,7 +171,7 @@ private static File extractLibraryFile(final String libFolderForCurrentOS, final
             // Check whether the contents are properly copied from the resource
             // folder
             try (InputStream nativeInputStream = NativeCodeLoader.class.getResourceAsStream(nativeLibraryFilePath)) {
-                try (InputStream extractedLibIn = new FileInputStream(extractedLibFile)) {
+                try (InputStream extractedLibIn = Files.newInputStream(Paths.get(extractedLibFile.getPath()))) {
```
garydgregory, line 174: See line 148 above, already a path I think.

garydgregory (changes requested):

src/main/java/org/apache/commons/crypto/stream/CryptoInputStream.java
arturobernalg: Ok, changed

src/main/java/org/apache/commons/crypto/NativeCodeLoader.java
arturobernalg: well seen. changed

ecki: Do you habe tested this for speed? I saw (and I don’t know why) performance regressions with the stream of Files.newInputStream compared to FileInputStream. Maybe that should be a separate change to introduce Path/NIO?

arturobernalg: > Do you habe tested this for speed? I saw (and I don’t know why) performance regressions with the stream of Files.newInputStream compared to FileInputStream. Maybe that should be a separate change to introduce Path/NIO?

Hi,
There is High Priority warning that recomend avoid FileStream. Cause garbage collection pauses

garydgregory: What warning? Issued by whom?

arturobernalg: https://pmd.github.io/latest/pmd_rules_java_performance.html#avoidfilestream
https://bugs.openjdk.java.net/browse/JDK-8080225

garydgregory: @arturobernalg 
Thanks for the pointers.

PR #129, opened 2021-01-08
Minor Improvements:

arturobernalg:
* collapsed Same Catch
* update jmh version to 1.27 
* add final var
* Add benchmark dependency globally 
* Add missing/complete javadoc
* PMD:
  ** Use ConcurrentHashMap
  ** Avoid FileStream

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,7 @@ The following provides more details on the included cryptographic software:
     User is able to specify the platform by maven profiles.-->
     <target.name>all</target.name>
     <maven-antrun-plugin.version>1.8</maven-antrun-plugin.version>
+    <jmh.version>1.27</jmh.version>
     <junit.version>5.7.0</junit.version>
 
     <!-- Override default buildNumber timestamp format, needed for coveralls plugin -->
@@ -365,7 +366,6 @@ The following provides more details on the included cryptographic software:
       <properties>
         <skipTests>true</skipTests>
         <benchmark>org.apache</benchmark>
-        <jmh.version>1.12</jmh.version>
       </properties>
 
       <dependencies>
@@ -729,6 +729,19 @@ The following provides more details on the included cryptographic software:
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- benchmark -->
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <version>${jmh.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <version>${jmh.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/src/main/java/org/apache/commons/crypto/cipher/JceCipher.java
+++ b/src/main/java/org/apache/commons/crypto/cipher/JceCipher.java
@@ -35,6 +35,9 @@ import javax.crypto.ShortBufferException;
  * Implements the {@link CryptoCipher} using JCE provider.
  */
 class JceCipher implements CryptoCipher {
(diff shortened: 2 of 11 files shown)


src/main/java/org/apache/commons/crypto/random/OsCryptoRandom.java
```diff
@@ -74,7 +76,7 @@ public OsCryptoRandom(final Properties props) {
 
         try {
             close();
-            this.stream = new FileInputStream(randomDevFile);
+            this.stream = Files.newInputStream(Paths.get(randomDevFile.getPath()));
```
arturobernalg, line 79: I'm not 100% sure about this change.

src/main/java/org/apache/commons/crypto/utils/ReflectionUtils.java
```diff
@@ -30,7 +32,7 @@
  */
 public final class ReflectionUtils {
 
-    private static final Map<ClassLoader, Map<String, WeakReference<Class<?>>>> CACHE_CLASSES = new WeakHashMap<>();
+    private static final ConcurrentMap<ClassLoader, Map<String, WeakReference<Class<?>>>> CACHE_CLASSES = new ConcurrentHashMap<>();
```
garydgregory, line 35: Why?

src/main/java/org/apache/commons/crypto/cipher/JceCipher.java
```diff
@@ -35,6 +35,9 @@
  * Implements the {@link CryptoCipher} using JCE provider.
  */
 class JceCipher implements CryptoCipher {
+    /**
+     *  The Cipher {@link Cipher} object.
```
garydgregory, line 39: Double word.

src/test/java/org/apache/commons/crypto/stream/AbstractCipherStreamTest.java
```diff
@@ -60,11 +60,11 @@
...
-    protected byte[] key = new byte[16];
-    protected byte[] iv = new byte[16];
-    protected int count = 10000;
-    protected static int defaultBufferSize = 8192;
-    protected static int smallBufferSize = 1024;
+    protected final byte[] key = new byte[16];
+    protected final byte[] iv = new byte[16];
+    protected final int count = 10000;
+    protected static final int defaultBufferSize = 8192;
+    protected static final int smallBufferSize = 1024;
```
garydgregory, line 67: Don't break binary compatibility.

pom.xml
```diff
@@ -731,5 +731,18 @@ The following provides more details on the included cryptographic software:
...
+      <artifactId>jmh-core</artifactId>
+      <version>${jmh.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <version>${jmh.version}</version>
+      <scope>test</scope>
+    </dependency>
```
garydgregory, line 746: Why add this if there are no benchmarks in this PR?

garydgregory (changes requested):
Breaks BC, see scattered comments.

arturobernalg (commented):
Resolved all the change that break the binary.

src/main/java/org/apache/commons/crypto/utils/ReflectionUtils.java
garydgregory: I don't think you've thought through the consequences of this change.

src/main/java/org/apache/commons/crypto/random/OsCryptoRandom.java
garydgregory: This is coded in a pretty obtuse manner, if the point is to use NIO, then start with a Path instead of a File.

In general, it feels to me like you need to step back and look at the big picture instead of applying whatever micro changes your IDE offers (or PMD, and so on).

arturobernalg: OK @garydgregory 
I close the PR and I will review the code
TY.

PR #133, opened 2021-02-27
Add isEmpty method check if an String is empty or null.

arturobernalg:


garydgregory:
@arturobernalg I implemented this differently in git master. Please review.

PR #161, opened 2022-06-09
CRYPTO-161 Add SM4 encryption support

wenwj0:
add sm4-ctr and sm4-cbc in opensslCipher

### DIFF
--- a/src/main/java/org/apache/commons/crypto/cipher/OpenSsl.java
+++ b/src/main/java/org/apache/commons/crypto/cipher/OpenSsl.java
@@ -45,7 +45,7 @@ final class OpenSsl {
 
     /** Currently only support AES/CTR/NoPadding. */
     private enum AlgorithmMode {
-        AES_CTR, AES_CBC, AES_GCM;
+        AES_CTR, AES_CBC, AES_GCM, SM4_CTR, SM4_CBC;
 
         /**
          * Gets the mode.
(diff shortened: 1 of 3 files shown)


jochenw:
Hi,

I am interested in the work, that you are doing here. Question: I notice, that your patch provides changes in the native code. That makes me hesitant: Before I accept the PR, could you help me to setup a build environment, where I can test these changes.

Background: We are considering a new release. As of now, there are no changes in the native code, so we could (in theory) just copy the native files from the previous release. But, for a release, that includes this PR, we would need a running build environment, where we can produce the updated files.

If this is interesting to you, please contact me privately at jochen@apache.org.

Thanks,

Jochen

garydgregory:
https://issues.apache.org/jira/browse/CRYPTO-161

garydgregory:
We need to be able to tell GitHub builds that we need OpenSSL >= 1.1.1

wenwj0:
I found a warning in [Java CI / build (macos-latest, 8, false)](https://github.com/apache/commons-crypto/runs/6812108742?check_suite_focus=true): `** WARN: Could not enable JNA; detected lib VERSION 0x10000000: Error looking up function 'ENGINE_load_rdrand': dlsym(0x7ff81e7bd070, ENGINE_load_rdrand): symbol not found`

Maybe the OpenSSL in Github builds is 1.0.0

PR #164, opened 2022-06-13
CRYPTO-162 openSslCipher support engine

wenwj0:
https://issues.apache.org/jira/projects/CRYPTO/issues/CRYPTO-162


garydgregory:
Please rebase on git master.

jochenw:
My impression is, that you wish to extend Crypto from a pure random generator towards a decoding/encoding engine.

While this seems interesting, I am nevertheless opposed to it: The native parts of Crypto are already a heavy maintenance burden, as they are. Mostly, due to the need to support multiple binary platforms. (Most of us have typically a single one, or two at the most.)

So, sorry!

wenwj0:
Thanks for replying.
I know that it's diffcult to maintain the native parts, I understand your choice. 
Besides, if you need any help, i can do it :)

garydgregory:
Since Crypto is a thin wrapper over OpenSSL, I don't see why we would not wrap more of its APIs, so this seems ok to me, if there were suitable tests.

PR #242, opened 2023-08-27
Set MacOS Runner to 11 and Fix Unknown Method Error

aremily:
Did a fork and build to get reacclimatized to the code base.  A few observations.  One, the build contained an unknown method error.  Two, the test matrix was pulling macos-latest, which used OpenSSL3 and broke the tests.  Three, the CI pipeline isn't running the full set of tests.  This PR addresses the first two issues.  I'm not sure what the rationale was for running only a small subset of tests in GitHub actions, but we may want to consider running them all, at least in some of the matrix builds.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -35,7 +35,7 @@ jobs:
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


aremily:
Including the console output for the unit tests.

[test_output.txt](https://github.com/apache/commons-crypto/files/12448334/test_output.txt)

PR #265, opened 2023-10-31
Bump to Maven 3.8.8

luhenry:
Maven 3.8.7 is not available for download anymore, so the previous URL would return a 404


garydgregory:
Why not 3.9.5?

luhenry:
> Why not 3.9.5?
It's a patch version bump, figured it would be easiest. But we should be able to do 3.9.5 indeed.

